PR #38, opened 2016-12-08
disable adding tasks to a completed job

paselem:
Don't let users add tasks to a job with an invalid state.


app/components/base/buttons/buttons.component.ts
```diff
@@ -50,6 +50,32 @@ export class AddButtonComponent {
...
+    public set title(value: string) {
+        this._title = value;
+    }
+    public get title() {
+        return this._title ? this._title : "Add";
+    }
+
+    public get enabled() {
+        return this.job
+            && this.job.state !== <any>"completed"
```
timotheeguerin: why this `<any>` there and below?
Should probably use the JobState enum instead

paselem: copy/paste :). Updated the rest of the file.

PR #43, opened 2016-12-12
Tests/job tests start

ascobie:
Testing job actions.


app/utils/constants.ts
```diff
@@ -31,4 +31,17 @@ export const forms = {
             priority: { min: -1000, max: 1000 },
         },
     },
+    validators: {
+        required: "required",
+        maxlength: "maxlength",
+        range: "validateRange",
+        pattern: "pattern",
+    },
+    fieldNames: {
```
timotheeguerin: I am not sure it really makes sense to use those.

Better of I think just do `form.controls.id` instead of `form.controls[fieldnames.id]`

we are not using it to define the control so it just make it tedious for testing and doesn't really prevent typos

PR #97, opened 2017-01-11
Heatmap v1

### LINKED ISSUE #70: Heatmap v1
Add a heatmap to the pool UI.

Work in [feature/heatmap](/Azure/BatchLabs/tree/feature/heatmap)

timotheeguerin:
fix #70


app/app.module.ts
```diff
@@ -104,6 +105,7 @@ import {
     commands,
 } from "./services";
 
+const modules = [PoolGraphsModule];
```
timotheeguerin: no the point is is should be an array containing all the sub modules we have in the app(This happens to be the first) It was getting out of hand to have this massive app.module file

app/assets/styles/main.scss
```diff
@@ -30,6 +30,9 @@
 @import "./pool/base";
 @import "./pool/create";
 
+// Node styles
+@import "./node/heatmap";
```
timotheeguerin: yep will change that

app/components/pool/graphs/nodes-heatmap.component.ts
```diff
@@ -0,0 +1,202 @@
...
+
+interface HeatmapTile {
+    index: number;
+    node: Node;
+}
+
+const stateTree: StateTree = [
+    { state: NodeState.idle, color: "#6ba3cb" },
+    { state: NodeState.running, color: "#388e3c" },
+    { state: NodeState.waitingForStartTask, color: "#94bdd9" },
```
timotheeguerin: Fixed

PR #117, opened 2017-01-20
Task detect errors and rerun task

### LINKED ISSUE #108: Rerun task
- [x] Rerun task with same setting (Reactivate api)
- [x] Rerun task width different settings(This needs to delete it and recreate it to keep the same id)

### LINKED ISSUE #109: Detect task error and propose quick fix
When a task fail propose to fix with 
- Rerun
- Rerun with different params(Clone)

timotheeguerin:
fix #108 
fix #109

### DIFF
--- a/app/assets/styles/base/banner.scss
+++ b/app/assets/styles/base/banner.scss
@@ -1,89 +1,131 @@
 bex-banner {
     display: block;
     margin-top: 10px;
+}
 
-    > md-card {
-        color: $whiteSmoke-darker;
-        background: $red;
-        padding: 0;
+bex-banner > md-card {
+    color: $whiteSmoke-darker;
+    padding: 0;
 
-        > .summary-container {
-            font-weight: 500;
+    > .summary-container {
+        font-weight: 500;
+        display: flex;
+        align-items: center;
+        height: 36px;
+
+        &.expandable {
+            cursor: pointer;
+        }
+
+        > [code] {
+            padding: 5px 24px;
+            height: 100%;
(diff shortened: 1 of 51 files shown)


app/assets/styles/base/banner.scss
```diff
@@ -1,112 +1,144 @@
...
 
         > .details-container {
-            padding: 10px 24px;
-            border-top: 1px solid $red-dark;
-            font-weight: 500;
+            border-top: 1px solid $accent;
         }
     }
+
+    .other-fixes {
```
paselem: What is the purpose of this class? Naming is a bit strange.

app/components/file/details/file-details.module.ts
```diff
@@ -0,0 +1,19 @@
+import { NgModule } from "@angular/core";
+
+import { commonModules } from "app/common";
+import { NodeBrowseModule } from "app/components/node/browse";
+import { PoolGraphsModule } from "app/components/pool/graphs";
```
paselem: Why do we need the pools Graphs module here? Same goes for the start task module.

app/assets/styles/base/banner.scss
timotheeguerin: In the banner there is the primary fix which is displayed directlry and then other fixes shown only after over

app/components/file/details/file-details.module.ts
timotheeguerin: yeah don't need those, might have been a copy and paste

PR #118, opened 2017-01-20
Feature/files

paselem:


app/components/file/browse/display/file-list-display-file-type-filter.pipe.ts
```diff
@@ -0,0 +1,20 @@
+import { Pipe, PipeTransform } from "@angular/core";
```
timotheeguerin: You used a pipe to filter those? I think it would be better to do a server side filter as we can do file startwtih applications, startup,etc.
How does this pipe work if the first batch of file doesn't actually contains some of those categories

app/components/file/browse/display/file-list-display.html
```diff
@@ -1,37 +1,32 @@
...
-    <!--<bex-quick-list-item *ngFor="let node of nodes" [routerLink]="['/pools', poolId, 'nodes', node.id]" routerLinkActive="selected">-->
-    <!--<bex-quick-list-item *ngFor="let node of nodes" [routerLink]="['/pools', poolId, 'nodes', node.id]" routerLinkActive="selected">
-        <h4 bex-quick-list-item-title>{{node.id}}</h4>
-        <p bex-quick-list-item-field>{{node.state}}</p>
-    </bex-quick-list-item>-->
-    <h1>TODO: Implement quicklist</h1>
-</bex-quick-list>
+<bex-focus-section>
+    <bex-quick-list *ngIf="quickList">
+        <bex-quick-list-item *ngFor="let file of files| fileDirectoryFilter" [routerLink]="['/files', file.url]"
```
timotheeguerin: What might be nice to add for the quick list/table is some sort of icons for files as editor do(This is a text file, this is a pyton file,etc.)

app/components/file/details/file-content.component.ts
```diff
@@ -63,49 +69,100 @@ export class FileContentComponent implements OnChanges, AfterViewInit {
...
 
     private _updateFileContent() {
-        this.currentSubscription = this.fileService.getFilePropertiesFromTask(this.jobId, this.taskId, this.filename)
-            .subscribe({
-                next: (result: any) => {
-                    if (result && result.data && result.data.properties) {
-                        const newContentLength = result.data.properties.contentLength;
-                        if (newContentLength !== this.lastContentLength) {
-                            this._loadUpTo(newContentLength);
+        // TODO: Refactor code!
```
timotheeguerin: TODO? agree this could use some cleanup

app/components/file/details/file-content.component.ts
```diff
@@ -63,49 +69,100 @@ export class FileContentComponent implements OnChanges, AfterViewInit {
...
-            if (this.followingLog) {
-                setTimeout(() => {
-                    this._scrollToBottom();
-                });
-            }
-            this.currentSubscription = null;
-        }, (e) => {
-            this._processError(e);
-        });
+        // TODO: Refactor code!
```
timotheeguerin: Same

app/components/file/details/file-details.component.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public downloadFile() {
+        // TODO: Download file
+        return;
+    }
+
+    // TODO: Add unit tests!
+    // TODO: Move to common utils
+    private parseRelativePath(fileUrl: string): any {
```
timotheeguerin: Bunch of todos here

app/components/file/home/file-home.component.ts
```diff
@@ -0,0 +1,78 @@
+import { Component, HostBinding, OnDestroy, OnInit } from "@angular/core";
+import { ActivatedRoute } from "@angular/router";
+import { Subscription } from "rxjs";
+
+import { SidebarManager } from "../../base/sidebar";
+
+@Component({
+    selector: "bex-file-home",
+    templateUrl: "./file-home.html",
```
timotheeguerin: you can remove this './' in front

app/components/file/details/file-details.component.ts
```diff
@@ -0,0 +1,93 @@
+import { Component, OnDestroy, OnInit } from "@angular/core";
+import { ActivatedRoute } from "@angular/router";
+import { autobind } from "core-decorators";
+import { Subscription } from "rxjs";
+
+import { FileService } from "app/services";
+
+@Component({
+    selector: "bex-file-details",
+    templateUrl: "./file-details.html",
```
timotheeguerin: './' not needed

app/components/file/home/file-home.component.ts
```diff
@@ -0,0 +1,78 @@
...
+        private sidebarManager: SidebarManager,
+        private activatedRoute: ActivatedRoute) {
+    }
+
+    public ngOnInit() {
+        this._paramsSubscriber = this.activatedRoute.params.subscribe((params) => {
+            this.url = params["url"];
+            let obj: any = this.parseRelativePath(this.url);
+
+            this.isJob = obj.type === "job";
```
timotheeguerin: Use enum or String enum here

app/services/file-service.ts
```diff
@@ -51,6 +51,44 @@ export class FileService extends ServiceBase {
...
+        poolId: string,
+        nodeId: string,
+        filename: string,
+        options: any = {}): Observable<FileContentResult> {
+
+        return Observable.fromPromise(BatchClient.file.getComputeNodeFile(poolId, nodeId, filename, options));
+    }
+
+    // TODO change to RxEntityProxy
+    public getFilePropertiesFromComputeNode(
```
timotheeguerin: RxEntityProxy? Add also a getOnce() which is usefull in many scenarios

app/utils/filter-builder/filter.ts
```diff
@@ -14,6 +14,10 @@ export class Filter {
         return formats.filter.format(this.properties.map(x => x.toOData()).join(` ${this.connector.id} `));
     }
 
+    public toODataWithName(): string {
```
timotheeguerin: why this?

app/components/file/browse/display/file-list-display-file-type-filter.pipe.ts
paselem: That's a good question. I didn't run into that issue. I liked this approach because swapping between the tabs is instant since we only download once. A service side filter could also work BUT we'd have a similar issue w/ the first batch of files not actually containing any files...

app/components/file/details/file-content.component.ts
paselem: Done.

app/components/file/details/file-content.component.ts
paselem: Done.

app/components/file/details/file-details.component.ts
paselem: Done.

app/components/file/details/file-details.component.ts
paselem: Done.

app/components/file/home/file-home.component.ts
paselem: Done.

app/components/file/home/file-home.component.ts
paselem: Done.

app/utils/filter-builder/filter.ts
paselem: I was investigating something... forgot to delete. Removed.

app/utils/fileUrlUtils.ts
```diff
@@ -2,7 +2,7 @@ import { FileType } from "app/models";
 import { Constants } from "app/utils";
 
 export function parseRelativePath(fileUrl: string): FileType {
-    let parts: string[] = fileUrl.split(Constants.FileUrlStrings.Separator);
+    const parts: string[] = fileUrl.split(Constants.FileUrlStrings.Separator);
```
timotheeguerin: Don't we have a problem with `\\` vs `/` here?

app/components/file/browse/display/file-list-display.html
paselem: Yup, good idea! I'll add that in later though. I'm wondering what to do w/ file types that we don't know about??? Or do we try to support icons for everything? For example .csv file? A .mov or .avi file? It might get a bit unwieldy... That said, there might be a library for that already... :)

app/utils/index.ts
```diff
@@ -6,6 +6,9 @@ export * from "./object";
 export * from "./os";
 export * from "./secure";
 
+import * as fileUrlUtils from "./fileUrlUtils";
```
timotheeguerin: Should not need to do that here

app/components/file/browse/display/file-list-display.html
timotheeguerin: Font awesome has a few general file icons we could use(text, image, etc.) otherwise maybe some other library

app/utils/fileUrlUtils.ts
paselem: I tested with both windows and linux machines. Looks like the URL separator is always consistent.

app/utils/index.ts
paselem: Why not? I'm exporting several functions here, similar to Constants.

PR #122, opened 2017-01-24
Breadcrumb use navigation

### LINKED ISSUE #2: Breadcrumbs should use navigation
Breadcrumb should be relative to the current navigation.

* When clicking on one of the left nav it should reset. 
* Should maybe only show the first item of the list and last few in case it get too long (Job a > ... > Node xyz > file.txt)

timotheeguerin:
fix #2 

Still need to add a force add breadcrumb


app/assets/styles/base/breadcrumbs.scss
```diff
@@ -60,14 +60,22 @@ bex-breadcrumbs {
...
+            }
         }
 
         > div.crumb-label {
             color: $whiteSmoke-darker;
             font-style: italic;
             text-align: right;
             line-height: 0.7em;
-            font-size: 0.7em;
+            font-size: 9px;
```
paselem: Why are you changing from em to px?

app/components/base/breadcrumbs/breadcrumb.service.ts
```diff
@@ -0,0 +1,145 @@
...
+        const cleaned = this._cleanupCrumbs(breadcrumb);
+        this._crumbs.next(cleaned.concat([breadcrumb]));
+    }
+
+    /**
+     * Compare segments of 2 breadcrumb
+     * If the 2 breadcrumb don't have the same base return false
+     */
+    public compareSegments(first: Breadcrumb, second: Breadcrumb) {
+        const firstSegments = first.segments;
```
paselem: as a public method, you may want to check for undefined

app/components/base/breadcrumbs/breadcrumb.service.ts
```diff
@@ -0,0 +1,145 @@
...
+        const secondSegments = second.segments;
+
+        for (let i = 0; i < Math.min(firstSegments.length, secondSegments.length); i++) {
```
paselem: by "same base" what do you mean? Just the same first item, or their entire path as to be the same? Depending on the answer, you may consider an early-exit by checking the lengths of the segments first.

app/components/base/breadcrumbs/breadcrumb.service.ts
```diff
@@ -0,0 +1,145 @@
+import { Injectable } from "@angular/core";
+import { ActivatedRoute, NavigationEnd, PRIMARY_OUTLET, Router } from "@angular/router";
+import { BehaviorSubject, Observable } from "rxjs";
+
+import { Constants } from "app/utils";
+import { Breadcrumb, RouteComponent } from "./breadcrumb-models";
+
+function breadcrumbMethodMessage(componentName) {
```
paselem: Minor: may want to rename this to "breadcrumbMissingMethodMessage" or something similar.

app/components/base/breadcrumbs/breadcrumb.service.ts
```diff
@@ -0,0 +1,145 @@
...
+            return null;
+        }
+        const data = component.breadcrumb(params, queryParams);
+        return {
+            data,
+            segments: path,
+            componentName: component.name,
+            queryParams: queryParams,
+            params: params,
+            url: `/${path.join("/")}`,
```
paselem: "/" -> const?

app/assets/styles/base/breadcrumbs.scss
timotheeguerin: made the text smaller and I don't like actually using the non fixed number for font because sometimes you get blurry text

app/components/base/breadcrumbs/breadcrumb.service.ts
timotheeguerin: good idea

app/components/base/breadcrumbs/breadcrumb.service.ts
timotheeguerin: actualy means if one of the 2 segments is a subset of the other

app/components/base/breadcrumbs/breadcrumb.service.ts
timotheeguerin: guess we can remove

PR #128, opened 2017-01-26
read only task dependencies UX

ascobie:
need navigation linked up, but should be mostly good to go. 
fix #[115]


app/components/task/details/task-dependencies.component.ts
```diff
@@ -0,0 +1,145 @@
...
+                return new TaskDependency(id);
+            });
+
+            this.dependencies.next(this.dependencies.value.concat(currentPage));
+            this.taskService.getMultiple(this.jobId, taskIdSet).subscribe({
+                next: (response: any) => {
+                    if (response.data) {
+                        this._processGetMultipleResponse(response, currentPage);
+                    }
+                },
```
timotheeguerin: I would let the task service handle that. The getMultiple should return the TaskObjects not the raw data

app/services/task-service.ts
```diff
@@ -89,6 +95,25 @@ export class TaskService extends ServiceBase {
         return getOnceProxy(this.get(jobId, taskId, options));
     }
 
+    public getMultiple(jobId: string, taskIds: string[], properties?: string): Observable<Task[]> {
```
timotheeguerin: Here I think you can use the RxListProxy instead with the fetchAll() method instead

app/components/task/details/task-dependencies.component.ts
```diff
@@ -0,0 +1,145 @@
+import { Component, Input, OnDestroy } from "@angular/core";
+
+import { Task, TaskDependency } from "app/models";
+import { TaskService } from "app/services";
+import { BehaviorSubject } from "rxjs";
```
timotheeguerin: Move in the extrenal import group

app/components/task/details/task-dependencies.component.ts
ascobie: The task service does return the task object. Or do you mean the ".data" bit?

app/components/task/details/task-dependencies.component.ts
```diff
@@ -0,0 +1,145 @@
...
+                const found = response.data.filter(item => item.id === td.id);
+                if (found && found.length > 0) {
+                    td.state = found[0].state;
+                    const dependencies = found[0].dependsOn;
+                    if (dependencies) {
+                        const count = this._taskDependenciesCount(dependencies);
+
+                        /* tslint:disable:no-empty */
+                        if (count === 0) {
+                        } else if (count <= 2) {
```
timotheeguerin: Maybe move this in another function

app/services/task-service.ts
ascobie: I thought you said to return an observable with the task[] data in it rather than using the ListProxy

app/components/task/details/task-dependencies.component.ts
ascobie: aye?

app/components/task/details/task-dependencies.component.ts
```diff
@@ -0,0 +1,145 @@
...
+            pageData.forEach(td => {
+                const found = response.data.filter(item => item.id === td.id);
+                if (found && found.length > 0) {
+                    td.state = found[0].state;
+                    const dependencies = found[0].dependsOn;
+                    if (dependencies) {
+                        const count = this._taskDependenciesCount(dependencies);
+
+                        /* tslint:disable:no-empty */
+                        if (count === 0) {
```
timotheeguerin: Here maybe move the count === 0 to be in the next if so you don't have to suppress the warning

`if(count > 0 && count <= 2)`

app/components/task/details/task-dependencies.component.ts
ascobie: what, like ... getDependsOnText() ?

app/components/task/details/task-dependencies.component.ts
timotheeguerin: yeah the data and no I don't think is this the task model, it is just the raw task structure

app/components/task/details/task-dependencies.component.ts
ascobie: ok, just copied directly from the portal code :)

app/services/task-service.ts
timotheeguerin: Yeah don't return the list proxy, you can just use it the function

timotheeguerin: that way it will also update the cache

app/components/task/details/task-dependencies.component.ts
timotheeguerin: move 3 lines above `rxjs` is not an part of our app

app/components/task/details/task-dependencies.component.ts
timotheeguerin: yep

app/services/task-service.ts
ascobie: i don't want it to update the cache, or it will slam the existing task entities in there.

app/components/task/details/task-dependencies.component.ts
ascobie: oh, gotcha ...

PR #154, opened 2017-02-06
Task filter by exit code

### LINKED ISSUE #153: Task filter by exit code
#152 needs filter by exit code to list the task that have error.

timotheeguerin:
Fix #153

### DIFF
--- a/app/components/base/advanced-filter/list-filter-control.ts
+++ b/app/components/base/advanced-filter/list-filter-control.ts
@@ -1,7 +1,7 @@
 import { Component, Input } from "@angular/core";
 import { FormControl, FormGroup } from "@angular/forms";
 
-import * as FilterBuilder from "app/utils/filter-builder";
+import { Filter, and, none, or, prop } from "app/utils/filter-builder";
 import { AdvancedFilter } from "./advanced-filter";
 import { AdvancedFilterControlBase } from "./control-base";
 
@@ -10,11 +10,32 @@ export enum ListFilterType {
     Exclude,
 }
 
+export interface ListFilterControlConfig {
+    /**
+     * If set to true all entries will be converted to numbers or ignored.
+     */
+    number?: boolean;
+
+    /**
+     * Allow list to take range.
+     * Range need to be speparated by the rangeSeparator
+     */
+    allowRanges?: boolean;
+
+    /**
+     * @default "-"
+     */
+    rangeSeparator?: string;
+}
+
 export class ListFilterControl extends AdvancedFilterControlBase {
     private _valueControl = new FormControl("");
 
-    constructor(label: string, private _expectNumber = false) {
+    constructor(label: string, private _config: ListFilterControlConfig = {}) {
         super(label);
+        if (!this._config.rangeSeparator) {
+            this._config.rangeSeparator = "|";
+        }
     }
 
     public formGroup(): FormGroup {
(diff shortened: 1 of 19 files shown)


app/components/base/advanced-filter/list-filter-control.ts
```diff
@@ -0,0 +1,141 @@
...
+    rangeSeparator?: string;
+}
+
+export class ListFilterControl extends AdvancedFilterControlBase {
+    private _valueControl = new FormControl("");
+
+    constructor(label: string, private _config: ListFilterControlConfig = {}) {
+        super(label);
+        if (!this._config.rangeSeparator) {
+            this._config.rangeSeparator = "|";
```
paselem: doc above says the default is "-"

app/components/base/advanced-filter/list-filter-control.ts
```diff
@@ -0,0 +1,141 @@
...
+            });
+            return and(...ranges, prop(this.name).notInList(values.items));
+        }
+    }
+
+    public parseValue(value: string): { items: any[], ranges: any[][] } {
+        if (!value) {
+            return { items: [], ranges: [] };
+        }
+        let values = value.replace(/\s/g, "").split(",").filter(x => x !== "");
```
paselem: Great example of a place to add some docs. My guess is that you're removing whitespaces?

app/components/base/advanced-filter/list-filter-control.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        let {items, ranges} = this._extractRanges(values) as any;
+
+        if (this._config.number) {
+            items = this._parseNumbers(items);
+            ranges = this._parseRangeNumbers(ranges);
+        }
+        return {
```
paselem: May want to have an object definition here since this is a public method

app/components/base/advanced-filter/list-filter-control.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private _parseNumbers(list: string[]): number[] {
+        return list.map(x => Number(x)).filter(x => !isNaN(x));
+    }
+
+    private _parseRangeNumbers(ranges: string[][]): number[][] {
+        return ranges.map((range) => {
+            return this._parseNumbers(range);
+        }).filter(x => x.length === 2);
```
paselem: What happens here if x.length === 2?

app/components/base/advanced-filter/list-filter-control.ts
```diff
@@ -0,0 +1,141 @@
...
+        return {
+            items,
+            ranges,
+        };
+    }
+}
+
+@Component({
+    selector: "bex-adv-filter-list",
+    template: `
```
paselem: Is it worth putting this in it's own file? Do we have any guidelines on when to use a new file and when to have it inline w/ the component?

app/components/base/list-and-show-layout/list-and-show-layout.component.ts
```diff
@@ -75,7 +75,7 @@ export class ListAndShowLayoutComponent implements AfterViewInit {
             if (query === "") {
                 this.quickFilter = FilterBuilder.none();
             } else {
-                this.quickFilter = FilterBuilder.prop(this.quickSearchField).startswith(query);
+                this.quickFilter = FilterBuilder.prop(this.quickSearchField).startswith(query.replace(/\s/g, ""));
```
paselem: Should /\s/g be a const somewhere? Looks like we're using it in more than one place. (If we do make it a const, make sure to write unit tests everywhere it's used to make sure if it get changed it doesn't break any consumers)

app/components/base/advanced-filter/list-filter-control.ts
timotheeguerin: Keep only valid range If less than 2 means one of the 2 entry is not a number

app/components/base/advanced-filter/list-filter-control.ts
timotheeguerin: Moved to a new file, and yeah should add some guideline on how to write a component.
I guess if the template is more than 4-5 lines is should go in a file

app/components/base/list-and-show-layout/list-and-show-layout.component.ts
timotheeguerin: added `"awodawpod  ".clearWhitespace()`

PR #159, opened 2017-02-07
Job detect error and prefill filter with url

### LINKED ISSUE #152: Job detect error
Detect errors  on a job and show the user.
Scheduling error contains those errors

- [x] when job timeout
- [x] when job terminated because task failed
- [x] job got a sheduling error

### LINKED ISSUE #157: Ability to fill advanced filter with the url
Should be able to fill the url with that advanced filter to have some link to some query

timotheeguerin:
fix #152 Job detect error
fix #157 Prefill filter with url


app/components/job/details/error-display/job-error-display.html
```diff
@@ -0,0 +1,16 @@
+<bex-banner *ngIf="jobFailed" fixMessage="View failed tasks" [fix]="gotoFailedTask" type="warning">
+    <div message>Job was terminated because one task failed.</div>
```
ascobie: should we say that "one task failed", or just that "a task failed". The first option might be confusing to users if in fact more than one task failed.

app/components/job/details/job-details.module.ts
```diff
@@ -0,0 +1,31 @@
...
+import { TaskBrowseModule } from "app/components/task/browse";
+
+import { JobErrorDisplayComponent } from "./error-display";
+import { JobDetailsHomeComponent } from "./job-details-home.component";
+import { JobDetailsComponent } from "./job-details.component";
+import { JobEnvironmentSettingsComponent } from "./job-env-settings.component";
+import { JobMetadataComponent } from "./job-metadata.component";
+import { JobPropertiesComponent } from "./job-properties.component";
+
+const components = [
```
ascobie: am wondering if we should make this job.module rather than job details only. then each area (job, task, app, ..) can have a single module and that way we easily know where the modules are when adding new bits.

app/components/job/details/error-display/job-error-display.html
timotheeguerin: well as soon as a task failed it gets terminated so there should only be 1 task that cause the job to fail. But I agree should be changed

app/components/job/details/job-details.module.ts
timotheeguerin: I already did Details module for task and nodes and they are on their own already quite big so I think it is good like that

PR #176, opened 2017-02-10
Notification persistence

### LINKED ISSUE #49: Notification persistance
Some notification should be persisted.
After auto dismiss the notification should move to a tray where the user can go back and see

timotheeguerin:
fix #49


app/assets/styles/base/notifications.scss
```diff
@@ -102,6 +102,11 @@ bex-notification {
         background-color: white;
     }
 
+    &.warn {
+        color: $yellow-dark;
```
timotheeguerin: change to white

app/assets/styles/base/notifications.scss
```diff
@@ -13,7 +13,7 @@ bex-notification-container {
     margin: 30px 10px;
     width: 320px;
     position: fixed;
-    z-index: 9998;
+    z-index: 9900;
```
ascobie: if we start to get a few of these z-indexes, it might pay to parameterise them.

timotheeguerin: good idea, i'll do that

PR #182, opened 2017-02-15
Improve task output and fix ctrl/shift click 

### LINKED ISSUE #175: Task output improvement
- [x] Make output first tab
- [x] Have the stdout/stderr show file size in the label

timotheeguerin:
fix #175


app/components/base/abstract-list/abstract-list-item-base.ts
```diff
@@ -56,23 +56,14 @@ export class AbstractListItemBase implements OnInit {
...
-
     /**
      * If the current item is active(Router)
      */
     public get isFocused(): boolean {
         return this.list.focusedItem === this.key;
     }
 
-    @HostListener("click", ["$event"])
+    // @HostListener("click", ["$event"])
```
paselem: remove?

client/api/batch-client-proxy/fileProxy.ts
```diff
@@ -97,7 +97,7 @@ export default class FileProxy {
                         isDirectory: headers["ocp-batch-file-isdirectory"],
                         url: headers["ocp-batch-file-url"],
                         properties: {
-                            contentLength: headers["content-length"],
+                            contentLength: parseInt(headers["content-length"], 10),
```
paselem: Default radix is already 10. Any reason you added it in specifically?

timotheeguerin: It is actually not the default. Default is 10 unless it starts with `0x`. But there was a warning telling that I should provide an argument

PR #188, opened 2017-02-16
Update/material beta.2

timotheeguerin:


timotheeguerin:
There is with the test. Looks like something like a memory leak.

PR #208, opened 2017-02-21
Fix/rename model files

ascobie:
still requires job-action and task-exit-conditions to be imported separately.


app/models/job-constraints.ts
```diff
@@ -1,4 +1,4 @@
-import { Constraints } from "./constraints";
+import { Constraints } from "app/models";
```
timotheeguerin: this is the same as doing `from "./"` should keep it as it was before

app/models/job-execution-information.ts
```diff
@@ -1,5 +1,6 @@
 import { Record } from "immutable";
-import { SchedulingError } from "./scheduling-error";
+
+import { SchedulingError } from "app/models";
```
timotheeguerin: same here

PR #209, opened 2017-02-21
cleanup console.error and console.warn

ascobie:


app/components/file/details/file-content.component.ts
```diff
@@ -171,7 +171,8 @@ export class FileContentComponent implements OnChanges, AfterViewInit {
             this.notFound = true;
             return;
         }
-        console.error("[FileContent.component] Error is", e.status, e);
+
+        log.error("[FileContent.component] Error is", { status: e.status, error: e });
```
ascobie: unsure if this is ok, or i should have done [ e.status, e ] as an array of params

timotheeguerin: yeah what I'd do here is 
```typescript
log.error(`[FileContent.component] Error is ${e.status}`, e)
```

PR #252, opened 2017-03-21
Splash screen while login the user and retrieving access tokens

### LINKED ISSUE #217: Add a splash screen while the app is loading/silently login in
Currently it seems like the app takes a few seconds to appear as it is silently login.

timotheeguerin:
fix #217


app/services/adal/user-authorization.ts
```diff
@@ -1,10 +1,12 @@
-import { remote } from "electron";
+import * as electron from "electron";
```
paselem: Why did you change this? What additional objects are you using from "electron" in this class?

timotheeguerin: none actually it was just conflicting with the remote service name I inject

PR #256, opened 2017-03-22
job and task tab refactor, app modules refactor

ascobie:
change properties tab to be "configuration" and include metadata and environment settings
remove most component imports from the app.module and replace with individual area modules.


ascobie:
includes: https://github.com/Azure/BatchLabs/issues/142

PR #298, opened 2017-04-07
Fix bug with query cache loading items too litle at the time

### LINKED ISSUE #297: QueryCache saving next link make it ignore the new params
Repro step:
Set pool picker maxResults to `2`
Go to pool list it will now only load pools 2 by 2.

timotheeguerin:
This is done by changing the way the query cache works. It will now just be used to show the initial items but items will still be loaded in the background(This should also make sure the data is not out of date)
fix #297

### DIFF
--- a/app/components/job/details/job-progress-status/job-progress-status.component.ts
+++ b/app/components/job/details/job-progress-status/job-progress-status.component.ts
@@ -37,7 +37,7 @@ export class JobProgressStatusComponent implements OnChanges, OnDestroy {
     constructor(private poolService: PoolService, private nodeService: NodeService) {
         this.poolData = poolService.get(null);
         this.data = nodeService.list(null, {
-            maxResults: 1000,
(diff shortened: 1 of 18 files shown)


app/components/job/details/job-progress-status/job-progress-status.component.ts
```diff
@@ -37,7 +37,7 @@ export class JobProgressStatusComponent implements OnChanges, OnDestroy {
     constructor(private poolService: PoolService, private nodeService: NodeService) {
         this.poolData = poolService.get(null);
         this.data = nodeService.list(null, {
-            maxResults: 1000,
+            pageSize: 1000,
```
paselem: I don't think you need to set this since the default is 1000.

timotheeguerin: we actually override the default to be 50 for all call, so I need to force it back

Issue #107, opened 2017-01-18
Pool autoscale formula


timotheeguerin:
Pool autoscale formula builder
Simply the autoscale formula with a builder that take a few inputs such as
- Min nodes
- Max nodes
- Cpu usage
- Queue length

PR #300, opened 2017-04-10
Feature: Pool autoscale formula 

johnnyzhang82:
Issue #107 Added scale mode picker in create pool form and integrated with codemirror text editor. Please review.
scalemode1
scalemode2
scalemode3


app/app.component.ts
```diff
@@ -44,27 +45,28 @@ export class AppComponent implements AfterViewInit, OnInit {
...
-        this.adalService.init(adalConfig);
-        this.accountService.loadInitialData();
-
-        this.hasAccount = accountService.currentAccount.map((x) => Boolean(x));
-
-        Observable
-            .combineLatest(accountService.accountLoaded, settingsService.hasSettingsLoaded)
-            .subscribe((loadedArray) => {
-                this.isAppReady = loadedArray[0] && loadedArray[1];
+            this.autoscaleFormulaService.init();
```
timotheeguerin: Can you add a blank line here instead of indenting. The auto formatter will put it back

app/components/pool/action/add/pool-create-basic-dialog.component.ts
```diff
@@ -78,4 +78,33 @@ export class PoolCreateBasicDialogComponent extends DynamicForm<Pool, PoolCreate
     public formToDto(data: any): PoolCreateDto {
         return createPoolToData(data);
     }
+
+    public changeScaleModeTab(event) {
+        if (event) {
```
timotheeguerin: Can this really be null?

app/services/autoscale-formula.service.ts
```diff
@@ -0,0 +1,72 @@
+import { Injectable } from "@angular/core";
+import { AutoscaleFormula } from "app/models";
+import { Constants, log } from "app/utils";
+import * as storage from "electron-json-storage";
```
timotheeguerin: I made a service wrapping around `electron-json-storage` called `LocalFileStorage`. Use that instead it makes the testing easier and it is also cleaner.

app/assets/styles/pool/create.scss
```diff
@@ -188,7 +185,7 @@ bl-autoscale-formula-picker  {
             display:block;
             padding:0 2px;
             transform:matrix(0.75, 0, 0, 0.75, 0, 0);
```
timotheeguerin: Why do you needs this?

package.json
```diff
@@ -109,6 +110,7 @@
     "brace": "^0.10.0",
     "bunyan": "^1.8.4",
     "chart.js": "^2.5.0",
+    "codemirror": "^5.25.0",
```
timotheeguerin: added the package twice? Should only be in the normal dependencies not dev

app/utils/autoscale.ts
```diff
@@ -0,0 +1,40 @@
...
+const variables = [
+    "ActiveTasks", "CPUPercent", "CurrentDedicated", "DiskBytes",
+    "DiskReadBytes", "DiskReadOps", "DiskWriteBytes", "DiskWriteOps",
+    "FailedTasks", "MemoryBytes", "NetworkInBytes", "NetworkOutBytes",
+    "PendingTasks", "RunningTasks", "SampleNodeCount", "SucceededTasks",
+    "TargetDedicated", "$WallClockSeconds",
+];
+
+// tslint:disable-next-line:max-line-length
+const keywordRegex = /\$CPUPercent|\$WallClockSeconds|\$MemoryBytes|\$DiskBytes|\$DiskReadBytes\$DiskWriteBytes|\$DiskReadOps|\$DiskWriteOps|\$NetworkInBytes|\$NetworkOutBytes|\$SampleNodeCount|\$ActiveTasks|\$RunningTasks|\$PendingTasks|\$SucceededTasks|\$FailedTasks|\$CurrentDedicated|\$TargetDedicated/;
```
timotheeguerin: I don't know how the codemirror modes works but could you here build the regex from an array here instead. It would look cleaner I think.

johnnyzhang82: This is the part I don't really like. I will refactor this regex to work better with a customize function.

timotheeguerin: cool

app/models/forms/create-pool-model.ts
```diff
@@ -35,11 +39,21 @@ export function createPoolToData(output: CreatePoolModel): PoolCreateDto {
...
-        targetDedicated: Number(output.targetDedicated),
+        targetDedicated: output.targetDedicated,
+        enableAutoScale: output.enableAutoScale,
+        autoScaleFormula: output.autoScaleFormula,
+        autoScaleEvaluationInterval: moment.duration({ minutes: output.autoScaleEvaluationInterval }),
         maxTasksPerNode: Number(output.maxTasksPerNode),
         enableInterNodeCommunication: output.enableInterNodeCommunication,
     };
 
+    if (output.enableAutoScale) {
```
johnnyzhang82: I tried to follow convention below this code, anyway it's now updated.

app/app.ts
```diff
@@ -10,6 +10,8 @@ import "chart.js";
 import "hammerjs";
 import "./utils/extensions";
 
+import "codemirror/addon/hint/show-hint.css";
```
johnnyzhang82: Installed yarn dependencies.

app/utils/autoscale.ts
johnnyzhang82: I've updated this part to a function that check whether given variables, function and system functions started with current word. If so, we display a list of word in auto complete ddl. Please review.

app/services/autoscale-formula.service.ts
johnnyzhang82: I think we also have to update ssh-key service as well. That's code I originally cloned from.

app/models/autoscale-formula.ts
```diff
@@ -0,0 +1,26 @@
...
+    name: null,
+    value: null,
+});
+
+export interface AutoscaleFormulaAttributes {
+    id: string;
+    name: string;
+    value: string;
+}
+export class AutoscaleFormula extends AutoscaleFormulaRecord implements AutoscaleFormulaAttributes {
```
johnnyzhang82: Fixed

app/components/pool/action/add/pool-create-basic-dialog.component.ts
johnnyzhang82: Probably not, I am just afraid of null reference exception.

package.json
johnnyzhang82: Removed dev-dependencies cm.

app/assets/styles/pool/create.scss
johnnyzhang82: This is style of label in code editor section. Are there any class we can reuse in form? Because I only see <input/> placeholder attribute as a subtitle when blur, I cloned same style for autoscale editor.

app/assets/styles/pool/create.scss
```diff
@@ -171,3 +171,90 @@ bl-vm-size-picker {
         }
     }
 }
+
+bl-autoscale-formula-picker {
```
timotheeguerin: I just merged a new way to handle styles. So you can use angular2 `styleUrls` check the `tags.component.ts` for a sample or the `components.md` doc. This will make styles cleaner

PR #305, opened 2017-04-11
change form layout

paselem:


app/components/pool/action/add/pool-create-basic-dialog.html
```diff
@@ -26,20 +26,20 @@
             </div>
         </bl-form-section>
 
-        <bl-form-section name="Select an operating system configuration">
+        <bl-form-section name="Select an operating system configuration" subtitle="Choose the operating system that will be installed on all the nodes in the pool">
```
timotheeguerin: should maybe change that to be `title` instead of `name` then

paselem: yeah, agreed. I thought about that too. Will change.

PR #309, opened 2017-04-13
Feature: New account dashboard

### LINKED ISSUE #310: Improve account landing area by showing a quick access to pools, jobs and applications
The current account dashboard is a bit too simple

paselem:
fix #310 

screen shot 2017-04-12 at 10 13 25 pm

Adding a new Dashboard + quick links to jobs, pools and app packages


paselem:
screen shot 2017-04-12 at 10 13 25 pm

Adding a new Dashboard + quick links to jobs, pools and app packages

app/assets/styles/account/details.scss
```diff
@@ -5,19 +5,14 @@ bl-account-details {
 
         .link {
```
timotheeguerin: cloud maybe have a bit less padding for those tiles(And also maybe rename from `.link`

app/components/account/details/account-details.html
```diff
@@ -16,19 +16,81 @@
...
-                <i class="fa fa-tasks fa-lg"></i>
-                <div>Jobs</div>
+            <md-card class="link">
+                <md-card-header>
+                    <md-card-title><h2>Job status</h2></md-card-title>
+                </md-card-header>
+                <md-card-content class="account-card-content">
+                    <div>
+                        <bl-loading [status]="jobData.status | async">
+                            <p *ngIf="(jobData.items | async).size === 0 && !loading"
```
timotheeguerin: There should be a `bl-no-item` component you give the size and the status and it will know when to show the message

app/models/pool.ts
```diff
@@ -105,6 +105,43 @@ export class Pool extends PoolRecord {
...
         this.tags = ModelUtils.tagsFromMetadata(this.metadata);
     }
+
+    public get osName(): string {
+        if (this.cloudServiceConfiguration) {
+            let osName: string;
+            let osFamily = this.cloudServiceConfiguration.osFamily;
+
+            if (osFamily === 2) {
+                osName = "Windows Server 2008 R2 SP1";
```
timotheeguerin: We are using the same thing in the OS picker maybe we could share this structure

client/splash-screen/splash-screen.ts
```diff
@@ -24,8 +24,8 @@ export class SplashScreen extends UniqueWindow {
 
     protected createWindow() {
         const window = new BrowserWindow({
-            height: 300,
-            width: 300,
+            height: 340,
```
timotheeguerin: Changed the splash screen size? No need to move anything inside too?

app/models/pool.ts
```diff
@@ -105,6 +105,43 @@ export class Pool extends PoolRecord {
...
+            let osName = "Windows Server";
+            osName += this.virtualMachineConfiguration.imageReference.sku;
+
+            return osName;
+        }
+
+        return "Linux";
+    }
+
+    public get osIconName(): string {
```
timotheeguerin: some of those methods I would maybe set the value in the constructor instead because the pool is immutable that will improve perfromance when angular detect for changes it won't have to do complex operation

app/components/shared/main-navigation.html
```diff
@@ -1,5 +1,8 @@
 <ul>
     <li>
+        <a href="#" [routerLink]="['/accounts', selectedId]" routerLinkActive="active"><i class="fa fa-th fa-lg"></i><span>Dash</span></a>
```
timotheeguerin: there might be a problem here. In the image it doesn't show the dashboard as active

app/components/account/details/account-details.html
paselem: Done

client/splash-screen/splash-screen.ts
paselem: Nope. Still looks good.

app/models/pool.ts
paselem: Moved all of this logic to the pool.ts class and consumed it from poolDecorator.

app/components/shared/main-navigation.html
paselem: Fixed. Was an issue with the top nav bar not being consistent with sending the account name to the rest of the app.

app/components/shared/main-navigation.component.ts
```diff
@@ -1,8 +1,24 @@
...
+    public selectedId: string;
+    public selectedAccountAlias: string = "";
+
+    constructor(private accountService: AccountService) {
+        accountService.currentAccountId.subscribe((accountId) => {
+            if (accountId) {
+                this.selectedId = accountId;
+                this.selectedAccountAlias = accountService.getNameFromAccountId(accountId);
+            } else {
+                this.selectedAccountAlias = "No account selected!";
```
timotheeguerin: where is the alias being used?

paselem: Alias is used as the display name. AccountId is the full Resource Id e.g. /subscription/.../resourceGroup/.../batchAccount/foo

PR #341, opened 2017-04-20
fix edit applcation forms layout

### LINKED ISSUE #344: Application edit form update to new form system.
Application edit form was missed in the form refactor

ascobie:
fix #344

A couple of forms were missed in the form refactor to bl-complex-form. Added back in the actionColor and actionName properties that were missed off when simple form implemented a complex form.

Also a couple of CSS tweaks to add some spacing between buttons.

### DIFF
--- a/app/components/base/form/form-page/form-page.scss
+++ b/app/components/base/form/form-page/form-page.scss
@@ -28,7 +28,6 @@
             }
         }
 
-
         margin-bottom: 5px;
     }
 
@@ -43,7 +42,7 @@
 
         > .section-content {
             flex: 1;
-            padding: 5px;
+            padding: 0 5px;
             overflow: hidden;
         }
     }
--- a/app/components/base/form/form-page/form-page.scss
+++ b/app/components/base/form/form-page/form-page.scss
@@ -38,12 +38,12 @@
 
         > .section-bar {
             border-right: 1px solid rgba(0, 0, 0, 0.12);
-            margin: 10px 0 10px 15px;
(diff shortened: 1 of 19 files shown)


app/styles/base/forms.scss
```diff
@@ -296,3 +296,7 @@ input.search-input {
     border: none;
     outline: none;
 }
+
+bl-submit-btn {
```
timotheeguerin: can this be a bit more specific that every `bl-submit-btn`

ascobie: nah i think it works fine with the way we do forms. every form has a submit button (or 2) and then a cancel button which is just a <button />. I have yet to see an example of it adding padding where it shouldn't and if i do i will fix it. Though i could nest it under: ".form-buttons" which would narrow it down a little more.

timotheeguerin: it is just here it will be applied to every bl-submit-btn even if it is not in a form

ascobie: nested it under form-buttons selector. this will be better.

ascobie: nested under form-buttons selector.

app/components/base/form/form-page/form-page.scss
```diff
@@ -43,7 +43,7 @@
 
         > .section-content {
             flex: 1;
-            padding: 0 25px;
+            padding: 0 5px;
```
timotheeguerin: I think 5 looks way to compact. Why did you want to change it?

image

PR #362, opened 2017-05-10
New Record decorator to define models

### LINKED ISSUE #361: Investigate a cleaner way to defined immutable models.
Currently there is duplicate when defining models which is prone to errors.
Move to a decorator system like the DTO.

Requiremenent: 
* Defined using decorators
* Immutable
* Use the same as immutable.js Record

timotheeguerin:
fix #361


app/core/record/errors.ts
```diff
@@ -0,0 +1,17 @@
...
+export class RecordMissingExtendsError extends Error {
+    constructor(ctr: Function) {
+        super(`Class ${ctr.name} with the @Model Decorator should also extends the Record class`);
+    }
+}
+
+/**
+ * Execption to be thrown if the user tries to call setter of attribute.
+ */
+export class RecordSetAttributeError extends Error {
```
paselem: Can we make this a compile-time error?

app/models/pool.ts
```diff
@@ -9,36 +10,6 @@ import { StartTask } from "./start-task";
...
-    taskSchedulingPolicy: null,
-    url: null,
-    virtualMachineConfiguration: null,
-    vmSize: null,
-    startTask: null,
-    metadata: List([]),
-    userAccounts: List([]),
-});
-
 export interface PoolAttributes {
```
paselem: Are these the only attributes to expose on the pool?

timotheeguerin: those are the attributes returned by the server

app/core/record/errors.ts
timotheeguerin: you can also but you need to add a readonly attribute before each property

```ts
@Prop()
public readonly id: string;
``

File: app/models/pool.ts [hunk 3]
```diff
@@ -70,47 +41,70 @@ export interface PoolAttributes {
 /**
  * Class for displaying Batch pool information.
  */
-export class Pool extends PoolRecord {
+@Model()
+export class Pool extends Record<PoolAttributes> {
+    @Prop()
     public allocationState: string;
+    @Prop()
```
ascobie: Don't suppose you would be keen to add a space between the property and the next @Prop()? It makes it pretty hard to read when there are no spaces between properties. I am sure this should violate a style guide somewhere :) well it would in C# anyway.

PR #384, opened 2017-05-19
Fix: up dashboard tables and a few other things

### LINKED ISSUE #392: Fix up the text overflow in the dashboard tables. Css fixes.

ascobie:
Fix #392
- add item-status to the account quicklist to bring it in-line with all the others. 
- fix up the text overflow in the dashboard tables
- add message about auto storage in the dashboard application card
- add correct error handling in the account details application loader to handle no auto storage


app/styles/base/banner.scss
```diff
@@ -4,19 +4,25 @@ bl-banner {
 }
 
 bl-banner > md-card.mat-card {
```
timotheeguerin: can you use this as an opportunity to move the `banner.scss` file in the same folder as the component and have the import "./banner.scss" in the component

app/styles/base/table.scss
```diff
@@ -10,6 +10,7 @@ bl-table {
     table {
```
timotheeguerin: same for table.scss

timotheeguerin:
Can you also attach an issue to this an put in milestone 0.4.0 so it's easier to do the changelog

app/styles/base/banner.scss
ascobie: sure thing.

ascobie:
"Can you also attach an issue .. "

So create an issue and then attach it to this RP?

PR #387, opened 2017-05-19
Feature: load the dashboard on app load if an account is selected in the top nav

paselem:


app/components/account/details/account-default.component.ts
```diff
@@ -10,8 +12,21 @@ import { Component } from "@angular/core";
...
         return { name: "Accounts" };
     }
+
+    constructor(
+        private accountService: AccountService,
+        private router: Router) {
+    }
+
+    public ngOnInit(): void {
+        this.accountService.currentAccountId.subscribe((accountId) => {
```
timotheeguerin: save the subscription and unsubsribe in ngOnDestroy or it will keep calling this everytime you cahnge account

PR #406, opened 2017-05-26
Feature: View and update associated storage account

### LINKED ISSUE #385: link storage account to batch account directly in the pool
rather than redirecting the user to the portal.

timotheeguerin:
fix #385 

image
image


ascobie:
Does the user have the ability to remove a storage account as well? I.e. select (none), like in the portal?

timotheeguerin:
Nope cannot remove account for now, will add this. And yeah still need to update the links

PR #455, opened 2017-06-13
Feature: Load all tasks running on a pool

### LINKED ISSUE #454: Pool heatmap tasks cross join instead of recent tasks
Recent tasks is limited to 10 tasks for a node which means if you have maxTaskPerNode > 10 the heatmap will only show up to 10 tasks runnign on each node even if there is more.

Do a list of all the tasks running on a given pool and join it with the nodes.

timotheeguerin:
fix #454

### DIFF
--- a/app/models/task.ts
+++ b/app/models/task.ts
@@ -1,6 +1,7 @@
-import { Record } from "immutable";
+import { List } from "immutable";
 import * as moment from "moment";
 
+import { ListProp, Model, Prop, Record } from "app/core";
 import { AffinityInformation } from "./affinity-information";
 import { ApplicationPackageReference } from "./application-package-reference";
 import { ComputeNodeInformation } from "./compute-node-information";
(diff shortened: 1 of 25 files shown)


app/models/task.ts
```diff
@@ -12,68 +13,59 @@ import { TaskDependencies } from "./task-dependencies";
...
+    creationTime: Date;
+    state: TaskState;
+    stateTransitionTime: Date;
+    previousState: TaskState;
+    previousStateTransitionTime: Date;
+    commandLine: string;
+    runElevated: boolean;
+    exitConditions: TaskExitConditions;
+    resourceFiles: ResourceFile[];
+    environmentSettings: NameValuePair[];
```
ascobie: should this be List<NameValuePair>?

timotheeguerin: no this is the input(what is returned by the server) so must be the native javascript object

PR #463, opened 2017-06-15
add additional items to the pool config tab

### LINKED ISSUE #462: Add the rest of the pool items to the configuration tab
Show OS Family  name, not just the ID
Show app package references and certificate references
Show pool metadata
Add start task execution info to the node configuration

ascobie:
Fix #462


app/models/certificate-reference.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * A reference to an application package to be deployed to a compute nodes
+ */
+export class CertificateReference {
```
timotheeguerin: Use the @Model thing this will fail

app/components/pool/details/pool-configuration.component.ts
```diff
@@ -11,30 +12,23 @@ export class PoolConfigurationComponent {
     @Input()
     public set pool(value: Pool) {
         this._pool = value;
-        this.decorator = new PoolDecorator(this._pool);
+        this._refresh(this._pool);
```
timotheeguerin: Prefer doing this in the ngOnChanges, it is made for that

app/models/certificate-reference.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * A reference to an application package to be deployed to a compute nodes
+ */
+@Model()
+export class CertificateReference extends Record<CertificateReferenceAttributes> {
+    @Prop() public thumbprint: string;
+    @Prop() public thumbprintAlgorithm: string;
+    @Prop() public storeLocation: string;
+    @Prop() public storeName: string;
+    @Prop() public visibility: string[];
```
timotheeguerin: Needs to be ListProp here

```
@ListProp(String) public visibility: List<string> = List([]);
```

Don't forget to set the default value to be an empty list too that way we never have to check for null for list

app/models/network-configuration.ts
```diff
@@ -0,0 +1,12 @@
+import { Record } from "immutable";
+
+const NetworkConfigurationRecord = Record({
```
timotheeguerin: Switch to @Model and Record

app/models/start-task-info.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * Class for displaying Start Task execution info from a node.
+ */
+@Model()
+export class StartTaskInfo extends Record<StartTaskInfoAttributes> {
+    @Prop() public state: TaskState;
+    @Prop() public startTime?: Date;
+    @Prop() public endTime?: Date;
```
timotheeguerin: don't make some attribute here optional `?` only in the interface for the attributes

app/utils/date-utils.ts
```diff
@@ -57,4 +57,21 @@ export class DateUtils {
...
+     * Using the supplied start and end times, compute the duration.
+     * @example 9m 2s, or null if no start time supplied.
+     *
+     * @param startTime - optional start time
+     * @param endTime - optional end time
+     */
+    public static computeRuntime(startTime?: Date, endTime?: Date): string {
+        if (!startTime) {
+            return null;
+        } else {
```
timotheeguerin: no need to else here

app/components/pool/details/pool-configuration.component.ts
ascobie: cool, though they are both set on change are they not.

Issue #470, opened 2017-06-16
ListKeys API call returns different format for standard storage. 


ascobie:
Classic storage returns: { primaryKey: string; secondaryKey: string; }
Standard storage returns :
```json
{
   "keys":[
      {
         "keyName":"key1",
         "permissions":"Full",
         "value":"KEY VALUE"
      },
      {
         "keyName":"key2",
         "permissions":"Full",
         "value":"KEY VALUE"
      }
   ]
}
```
Change StorageClientService to handle both.

PR #471, opened 2017-06-16
fix bug with list storage key formats

ascobie:
Fix: #470


ascobie:
Think i messed this PR up. I based of off my feature/additional-pool-config-details branch, when i should have done it from Master. Need to merge this one in first, followed by the above.

PR #483, opened 2017-06-20
Test for windows offer types that are differnt format to existing

### LINKED ISSUE #481: New rendering images with incorrect icon in Labs
Created a pool with the new windows rendering image and the icon in batch labs is the linux icon. 

linux-pool-issue

ascobie:
Fix #481 

Hope this covers all cases now. 

Edit: Has broken the RDP UX with: 

_./app/components/node/connect/node-connect.component.ts:61:17
TS2322: Type 'boolean | RegExpMatchArray' is not assignable to type 'boolean'.
Type 'RegExpMatchArray' is not assignable to type 'boolean'._

I will fix it in the morning.

### DIFF
--- a/app/utils/pool-utils.ts
+++ b/app/utils/pool-utils.ts
@@ -58,8 +58,8 @@ export class PoolUtils {
         return !this.isOfferWindows(offer);
     }
 
-    public static isOfferWindows(offer: string) {
-        return Boolean(offer.match(/windows/i));
+    public static isOfferWindows(offer: string): boolean {
+        return /windows/i.test(offer);
     }
 
     public static iconForOffer(offerName: string) {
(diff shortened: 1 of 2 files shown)


ascobie:
The new offer is: rendering-windows2016 so this should work now.

app/utils/pool-utils.ts
```diff
@@ -59,7 +59,7 @@ export class PoolUtils {
     }
 
     public static isOfferWindows(offer: string) {
-        return /^.*Windows.*$/.test(offer);
```
timotheeguerin: to fix your problem do it the way it was but with the i for case insensitive

/windows/i.test(offer)

also mark the function to return a boolean explicitly

PR #505, opened 2017-06-22
Show correct icons for new rendering skus

ascobie:
Fix: #504


app/utils/pool-utils.ts
```diff
@@ -14,10 +14,12 @@ const iconMapping = {
     "SLES": Icons.openSUSE,
     "SLES-HPC": Icons.openSUSE,
     "standard-data-science-vm": Icons.windows,
+    "rendering-windows2016": Icons.windows,
```
timotheeguerin: I know this is not part of this pull request but I would move the rendering images to another tab so we have VMs, Data science Vms, and rendering vms

ascobie: Good point about that. I kind of missed the way these VM's were displayed and have only now seen the Data Science VM's tab. Makes perfect sense to have them in another tab. Will do that in this PR.

PR #517, opened 2017-06-27
Quick add file auto focus and close autocomplet on blur

### LINKED ISSUE #486: Quick add files to the output should auto focus
The search box is not focussed automatically
image

timotheeguerin:
fix #486


app/components/task/details/output/task-log.component.ts
```diff
@@ -41,14 +44,14 @@ export class TaskLogComponent implements OnInit, OnChanges, OnDestroy {
...
-        });
-   }
+    public ngOnInit() {
+        this.filteredOptions = this.filterControl.valueChanges
+            .map((nameFilter) => {
+                return nameFilter
+                    ? this._options.value.filter(option => new RegExp(`${nameFilter}`, "gi").test(option))
+                    : this._options.value;
+            });
+    }
```
ascobie: nitpick, is this an extra space added to this method?

timotheeguerin: yep seems like so, not sure why tslint is not catching that

PR #524, opened 2017-06-28
Poll file size for changes. Retry failed calls for file content.

### LINKED ISSUE #523: task log file sizes don't reload
There are 2 issues with the task log tab. 

1) the file sizes don't reload after the initial load.
2) should a call to read the file contents fail with either a 404 or 409, where the stdout/stderr file has not been created yet, or the task is not yet running on the node, the load is not retried.

ascobie:
Fix #523


ascobie:
waiting to find out why the API errors are not being returned in the correct format. i need this so i can catch and retry file content calls.

Issue #498, opened 2017-06-22
Add application licenses picker feature to Labs.


ascobie:

ascobie:
merged into master

PR #526, opened 2017-06-28
Application license picker for rendering images

ascobie:
Feature #498

license-picker


app/components/base/form/form-page/form-page.component.ts
```diff
@@ -57,7 +57,7 @@ export class FormPageComponent {
     public openedWith: FocusableElement;
 
     // tslint:disable-next-line:no-forward-ref
-    constructor( @Inject(forwardRef(() => ComplexFormComponent)) private form: ComplexFormComponent) { }
```
timotheeguerin: I think the formatter adds the space can you check it does that for you so we don't change that back on forth

app/components/base/table/table.scss
```diff
@@ -31,7 +31,7 @@ bl-table {
         display: table-header-group;
 
         > tr {
-            cursor:pointer;
```
timotheeguerin: hhm without the pointer does it still look alright for other tables? Maybe we should have a class for table that are selcatble or not and show the pointer then

app/components/pool/action/add/app-license-picker.component.ts
```diff
@@ -0,0 +1,128 @@
...
+})
+export class AppLicensePickerComponent implements ControlValueAccessor, OnInit, OnChanges, OnDestroy, Validator {
+    @ViewChild("licenseTable")
+    public table: TableComponent;
+
+    public licenses: List<ApplicationLicense> = List([]);
+
+    private _propagateChange: (value: string[]) => void = null;
+    private _propagateTouched: (value: boolean) => void = null;
+    private _pickedLicenses: Map<string, boolean> = Map<string, any>({});
```
timotheeguerin: `= Map<string, boolean>`

app/components/pool/action/add/app-license-picker.component.ts
```diff
@@ -0,0 +1,128 @@
...
+            new ApplicationLicense({
+                id: "arnold",
+                description: "Autodesk Arnold",
+                licenseAgreement: "",
+                cost: "2c USD",
+            }),
+        ]);
+    }
+
+    public ngOnChanges(inputs) {
```
timotheeguerin: remove those if not used I think

app/components/pool/action/add/app-license-picker.component.ts
```diff
@@ -0,0 +1,128 @@
...
+    public viewEula(name: string) {
+        let config = new MdDialogConfig();
+        config.viewContainerRef = this.viewContainerRef;
+        const dialogRef = this.dialog.open(LicenseEulaDialogComponent, config);
+        dialogRef.componentInstance.licenseName = name;
+    }
+
+    public eulaCheck(event: MdCheckboxChange) {
+        this._eulaRead = event.checked;
+        if (this._propagateChange) {
```
timotheeguerin: make this if() as an funciton as you use it twice

app/styles/variables.scss
```diff
@@ -51,6 +51,7 @@ $link-color                 : map-get($primary, 700);
 $link-color-hover           : map-get($primary, 800);
 $validation-error-color     : map-get($danger, 500);
 $label-text                 : $dusty-grey;
+$portal-link-color          : #00bcf2;
```
timotheeguerin: hhm can you make this depends on one of $primary, $danger, $warn, etc

app/components/pool/action/add/app-license-picker.component.ts
ascobie: good catch ...

app/components/pool/action/add/app-license-picker.component.ts
ascobie: yeah i left them there as this will change as soon as we have an API to call. easy enough to add back in so no worries.

app/components/base/table/table.scss
ascobie: ill push it like this as i think its ok and in-line with tables in the portal. though if we decide otherwise we can add a specific style to this component.

PR #564, opened 2017-07-26
quick fix to get pools working

ascobie:
am sure it can be done better, but this at least allows the UI to work.


app/components/pool/details/pool-configuration.html
```diff
@@ -57,7 +57,7 @@
             <bl-text-property label="User identity" [value]="decorator.userIdentitySummary"></bl-text-property>
```
timotheeguerin: I would wrap all the start task properties into a div or template and have a ngIf startTask instead

PR #598, opened 2017-08-08
NCJ UX with Mode selection

t-anjam:
I added mode selection option to the Market feature in BatchLabs. Now users can create pools, job + old pool, job + new pool submissions.
image
image
image
image
image
image


src/client/python-process/python-executable.ts
```diff
@@ -41,7 +41,7 @@ export function tryPython(pythonPath: string): Promise<string> {
  * @returns Promise that resolve the first valid path. If no path are valid reject with an error.
  */
 export function tryMultiplePythons(paths: string[]): Promise<string> {
-    let promise: Promise<string> = Promise.reject(null);
+    //let promise: Promise<string> = Promise.reject(null);
```
timotheeguerin: yeah just remove that line

python/controllers/ncj_controller.py
```diff
@@ -7,13 +7,13 @@
 
 @app.procedure("submit-ncj-job")
 def submitNcjJob(request: JsonRpcRequest, template, parameters):
```
timotheeguerin: was like that before but should follow python coding style so `submit_ncj_job`

python/controllers/ncj_controller.py
```diff
@@ -26,3 +26,10 @@ def createNcjPool(request: JsonRpcRequest, template, parameters):
     print("Poolparm from json is done")
     client.pool.add(pool)
     return {"what": "it works"}
+
+@app.procedure("get-ncj-pool")
+def getNcjPool(request: JsonRpcRequest, template, parameters):
```
timotheeguerin: can you call that `expand_ncj_pool` instead(and change at id also to be `expand-ncj-pool`

app/components/market/application-action/choose-action.component.ts
```diff
@@ -22,7 +22,11 @@ export class ChooseActionComponent implements OnInit, OnDestroy {
 
     private _paramsSubscriber: Subscription;
 
-    constructor(private templateService: NcjTemplateService, private route: ActivatedRoute) { }
+    constructor(private templateService: NcjTemplateService, private route: ActivatedRoute) {
+
```
timotheeguerin: remove spaces as it was before

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
         private route: ActivatedRoute,
         private router: Router,
-        private templateService: NcjTemplateService) { }
-
-    public ngOnInit() {
+        private templateService: NcjTemplateService) {
+             this.mode_state = "None";
+             this.form = new FormGroup({a: new FormControl()});
+    }
+    private ngOnInit() {
```
timotheeguerin: ngOnInit is public

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
     constructor(
         public formBuilder: FormBuilder,
         private pythonRpcService: PythonRpcService,
         private route: ActivatedRoute,
         private router: Router,
-        private templateService: NcjTemplateService) { }
-
-    public ngOnInit() {
+        private templateService: NcjTemplateService) {
+             this.mode_state = "None";
```
timotheeguerin: Use an enum for this

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+             this.form = new FormGroup({a: new FormControl()});
```
timotheeguerin: `a:` ?

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+                this.poolTemplate = templates.pool;
+                this.title = `Run ${this.actionId} from ${this.applicationId}`;
+                this.createForms();
+            });
         });
     }
-
-    public createForms() {
-        let parameterKeys = Object.keys(this.jobTemplate.parameters);
+    private createForms() {
```
timotheeguerin: make private methods start with `_`

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
-        this.form = this.formBuilder.group({
-            job: this.jobFormGroup,
-            pool: this.pickedPool,
-        });
-        if (this._formValue) {
-            this.form.setValue(this._formValue);
+        for (let key of poolParameters) {
+            if ("defaultValue" in this.poolTemplate.parameters[key]) {
+                const defaultValue = String(this.poolTemplate.parameters[key].defaultValue);
+                fg[key] = new FormControl(defaultValue);
```
timotheeguerin: I think you should separate the pool and job properties in 2 different file group.

Form should be 
```ts
this.jobParams = new FormGroup(jobFg)
this.poolParams = new FormGroup(poolFg)

`this.form = formBuilder.group({pool: this.poolParams, job: this.jobParams});
``

File: app/components/market/submit/submit-market-application.html [hunk 10]
```diff
@@ -1,37 +1,223 @@
...
+                            <md-input-container class="form-element">
+                            <input mdInput type="text" name="name" [formControlName]="param" [placeholder]="param">
+                                <md-hint align="end">{{jobTemplate.parameters[param].metadata.description}}</md-hint>
+                            </md-input-container>
+                        </div>
+                        <div *ngSwitchCase="'file-group'">
+                            <bl-file-group-picker [formControlName]="param" [label]="param" class="form-element" [hint]="jobTemplate.parameters[param].metadata.description">
+                            </bl-file-group-picker>
+                        </div>
+                        <div *ngSwitchCase="'file-in-file-group'">
```
timotheeguerin: There is way to much duplicates here. You should probably create another component called `ncj-form-parameter` to which you give the param type and it will choose the right value

app/components/market/submit/submit-market-application.html
```diff
@@ -1,37 +1,223 @@
...
+                            <bl-cloud-file-picker [containerId]="getContainerFromFileGroup(form.controls[jobTemplate.parameters[param].metadata.dependsOn]?.value)" [formControlName]="param"
+                            [label]="param" class="form-element" [hint]="jobTemplate.parameters[param].metadata.description">
+                            </bl-cloud-file-picker>
+                        </div>
+                        <div *ngSwitchCase="'dropdown'">
+                            {{param}}
+                            <select [formControlName]="param">
```
timotheeguerin: use a md-select here too

src/client/python-process/python-executable.ts
t-anjam: done

python/controllers/ncj_controller.py
t-anjam: done

python/controllers/ncj_controller.py
t-anjam: done

app/components/market/application-action/choose-action.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: removed

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: let's wait till UX is more refined

app/components/market/submit/submit-market-application.html
t-anjam: let's wait till UX converges to a final design

app/components/market/submit/submit-market-application.html
t-anjam: done

app/components/market/submit/submit-market-application.html
timotheeguerin: what do you mean? you are going to need that either way. This looks way to hacky this way

app/components/market/submit/submit-market-application.component.ts
timotheeguerin: This will be much better that merging all params togther and then trying to figure out which ones are belong to the pool and which ones to the job

app/components/market/submit/submit-market-application.html
t-anjam: I was thinking it is better to wait in refactoring code until a finalized design is agreed upon. I predict there might be more changes to the UX in the nearby future.

timotheeguerin: yes but this is completely independent from whatever the UI will looks like. You just make a component that display the right input dependending on the type. This is something we should have either way(even if there was no duplicate).

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
-            if (params.formParams) {
-                try {
-                    const value = JSON.parse(params.formParams);
-                    this._formValue = value;
-                    this.form.setValue(value);
-                } catch (e) {
-                    log.warn("Invalid form param. Not valid json.", params.formParams as any);
-                }
-            }
+            this.templateService.getTemplates(this.applicationId, this.actionId).subscribe((templates) => {
```
paselem: minor: why is this 'templates' instead of 'template' (note plural vs. singular')?

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+    private createForms() {
+        let jobParameters = [];
+        let poolParameters = [];
+        if (this.jobTemplate && this.jobTemplate.parameters){
+            jobParameters = Object.keys(this.jobTemplate.parameters);
+        }
+        if (this.poolTemplate && this.poolTemplate.parameters){
+            poolParameters = Object.keys(this.poolTemplate.parameters);
+        }
         let fg = {};
```
paselem: not a big fan of abbreviated variables. I assume we can expand to formGroup?

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
-        for (let key of parameterKeys) {
+
+        for (let key of jobParameters) {
             if ("defaultValue" in this.jobTemplate.parameters[key]) {
```
paselem: shouldn't "defaultValue" be a const?

app/components/market/submit/submit-market-application.component.ts
paselem: @t-anjam , how will the UX changes affect this? Is there something that we need to discuss in the UX which will affect our decision to keep the job and pool sections of the template separate?

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
-    public submit() {
-        this.error = null;
-        const jobParams = this.jobFormGroup.value;
-        // RPC takes in Template JSON object and Parameter JSON object
-        const obs = this.pythonRpcService.callWithAuth("submit-ncj-job", [this._buildJobTemplate(), jobParams]);
+    private runJobWithPool(expandedPoolTemplate, jobInputs){
+        delete expandedPoolTemplate.id;
+        this.jobTemplate.job.properties.poolInfo = {
+            autoPoolSpecification: {
+            autoPoolIdPrefix: "jobname",
```
paselem: should the prefix be the actual job name, not the string "jobname"?

Also, consider what kinds of constraints you will run into here. Batch has limitations on what values are allowed for this. It would be worth taking a look at the Batch docs to know what issues you may run into: https://docs.microsoft.com/en-us/rest/api/batchservice/

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+            poolLifetimeOption: "job",
+            keepAlive: false,
+            pool: expandedPoolTemplate,
+            },
+        };
+        const obs = this.pythonRpcService.callWithAuth("submit-ncj-job", [this.jobTemplate, jobInputs]);
         obs.subscribe({
-            next: (data) => this._redirectToJob(data),
+            next: (data) => this._redirectToJob(),
```
paselem: Why did you remove the 'data' object from this call? The expected design is to navigate to the job we just created. Looking at the code below, you no longer seem to have a reference to the job id and can only navigate to the jobs list UI.

Also, a second comment. We should have a type on data, eg.
```ts
next: (data: myObj) => this._redirectToJob()
```

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
             error: (err) => this.error = ServerError.fromPython(err),
         });
-        return obs;
```
paselem: If you are no longer returning the obs, you can probably get rid of the variable and do it all in a 'single' line:

```ts
this.pythonRpcService.callWithAuth("...", [...]).subscribte({...});
```

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+        let poolInputs = {};
+        for (let param of Object.keys(this.form.controls)) {
+            if (this.getParameters(this.jobTemplate).includes(param)){
+                jobInputs[param] = this.form.controls[param].value;
+            }
+            else if (this.getParameters(this.poolTemplate).includes(param)) {
+                poolInputs[param] = this.form.controls[param].value;
+            }
+        }
+        if (this.mode_state === "PoolNJob"){
```
paselem: "PoolNJob" should be changed to a const or enum. Probalby worth renaming to something more obvious.

Same comment goes for "PoolOJob" and "Pool"

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+        }
+        else if (this.mode_state === "PoolOJob"){
+            this.jobTemplate.job.properties.poolInfo = this.pickedPool.value;
+            const obs = this.pythonRpcService.callWithAuth("submit-ncj-job", [this.jobTemplate, jobInputs]);
+            obs.subscribe({
+                next: (data) => this._redirectToJob(),
+                error: (err) => this.error = ServerError.fromPython(err),
+            });
+        }
+        else if (this.mode_state === "Pool"){
```
paselem: Instead of a long if/else if section, consider using a switch statement:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-1-8.html

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+        delete expandedPoolTemplate.id;
+        this.jobTemplate.job.properties.poolInfo = {
+            autoPoolSpecification: {
+            autoPoolIdPrefix: "jobname",
+            poolLifetimeOption: "job",
+            keepAlive: false,
+            pool: expandedPoolTemplate,
+            },
+        };
+        const obs = this.pythonRpcService.callWithAuth("submit-ncj-job", [this.jobTemplate, jobInputs]);
```
paselem: What is "submit-ncj-job" call out to?

Also, should be a constant.

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
-        const parameters = this.jobTemplate.parameters;
-        const otherParameters: any[] = [];
-        for (let name of Object.keys(parameters)) {
-            const param = parameters[name];
-            if (name === ConventionNames.jobName) {
-                this.jobNameParam = new NcjParameterWrapper(name, param);
-            } else {
-                otherParameters.push(new NcjParameterWrapper(name, param));
+    private getContainerFromFileGroup(fileGroup: string) {
+        return fileGroup && `fgrp-${fileGroup}`;
```
paselem: file group prefix 'fgrp-' should be a constant.

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+    }
+    private getType(param){
+        if (this.getParameters(this.jobTemplate).includes(param)){
+            if (this.jobTemplate.parameters[param].metadata.advancedType){
+                return this.jobTemplate.parameters[param].metadata.advancedType;
+            }
+            else if (this.jobTemplate.parameters[param].allowedValues){
+                return "dropdown";
```
paselem: "dropdown" and other types should be enums or constants.

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+            }
+            return this.jobTemplate.parameters[param].type;
```
paselem: Is there ever a case where the type param a user enters is invalid in a way which causes the ui to error/crash? What if a user has a typo in their template for example?

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
+        else if (this.getParameters(this.poolTemplate).includes(param)){
+            if (this.poolTemplate.parameters[param].metadata.advancedType){
+                return this.poolTemplate.parameters[param].metadata.advancedType;
             }
+            else if (this.poolTemplate.parameters[param].allowedValues){
+                return "dropdown";
+            }
+            return this.poolTemplate.parameters[param].type;
+        }
+        else {
```
paselem: I may be reading this method incorrectly, but should we also have a check for 'int'? I don't know if that would be considered and 'advancedType' but I wouldn't think so.

app/components/market/submit/submit-market-application.html
```diff
@@ -1,37 +1,223 @@
...
+                        <button style="margin:5px; width:200px" (click) = "mode_state = 'PoolNJob'" type="button" md-raised-button>new Pool + Job</button>
+                        <button style="margin:5px; width:200px" (click) = "mode_state = 'PoolOJob'" type="button" md-raised-button>existing Pool + Job</button>
+                        <button style="margin:5px; width:200px" (click) = "mode_state = 'Pool'" type="button" md-raised-button>new Pool</button>
+                    </ul>
+                </bl-form-section>
+            </bl-form-page>
+        </bl-complex-form>
+    </div>
+
+    <div *ngIf="mode_state == 'PoolNJob'">
```
paselem: Inconsistent spacing around operators throught the file. In some cases you are using 
```html
'a==b'
```
and in others
```html
'a == b'
```

paselem: There is also an inconsistent use of ' and ".

Is the linter not catching these?

app/components/market/submit/submit-market-application.html
```diff
@@ -1,37 +1,223 @@
...
-        <bl-button [action]="submit" type="wide" class="submit-job" color="success">
-            <i class="fa fa-play"></i>
-        </bl-button>
+                </bl-form-section>
+            </bl-form-page>
+        </bl-complex-form>
     </div>
+
+    <button *ngIf="mode_state!='None'" style="width:100px;background-color:#3e81b0;" (click) = "onSubmit()" type="button"  md-raised-button>Submit</button>
+
```
paselem: nitpick: lots of empty lines here.

python/controllers/ncj_controller.py
```diff
@@ -7,13 +7,13 @@
 
 @app.procedure("submit-ncj-job")
 def submitNcjJob(request: JsonRpcRequest, template, parameters):
+    print("Got template", template)
```
paselem: remove print statement.

python/controllers/ncj_controller.py
```diff
@@ -7,13 +7,13 @@
...
     client = request.auth.client
     job_json = client.job.expand_template(template, parameters)
-    print("Job json", job_json)
     job = client.job.jobparameter_from_json(job_json)
     print("Job is", job)
     client.job.add(job)
-    return job_json
+    return {"what": "it works"}
```
paselem: Was this just for debugging? Change it back?

python/controllers/ncj_controller.py
```diff
@@ -7,13 +7,13 @@
 
 @app.procedure("submit-ncj-job")
 def submitNcjJob(request: JsonRpcRequest, template, parameters):
+    print("Got template", template)
     client = request.auth.client
     job_json = client.job.expand_template(template, parameters)
-    print("Job json", job_json)
     job = client.job.jobparameter_from_json(job_json)
     print("Job is", job)
```
paselem: remove print statement.

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
-import { autobind } from "core-decorators";
-
-import { NcjJobTemplate, NcjParameter, ServerError } from "app/models";
+import { ServerError } from "app/models";
 import { NcjTemplateService, PythonRpcService } from "app/services";
-import { ObjectUtils, log } from "app/utils";
-import * as inflection from "inflection";
 import "./submit-market-application.scss";
 
-enum NcjParameterExtendedType {
```
timotheeguerin: This enum was there before you should keep it and add more types

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,171 @@
...
-    string = "string",
-    int = "int",
-    fileGroup = "file-group",
-    fileInFileGroup = "file-in-file-group",
-}
-
-class NcjParameterWrapper {
-    public type: NcjParameterExtendedType;
```
timotheeguerin: Same this class should still be here instead of the function

app/components/market/submit/submit-market-application.component.ts
timotheeguerin: it needs to return this observable for the button to show the loading actually

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -1,208 +1,173 @@
...
+        const obs = this.pythonRpcService.callWithAuth("submit-ncj-job", [this.jobTemplate, jobInputs]);
         obs.subscribe({
-            next: (data) => this._redirectToJob(data),
+            next: (data) => this._redirectToJob(),
             error: (err) => this.error = ServerError.fromPython(err),
         });
-        return obs;
     }
 
-    private _redirectToJob(data) {
```
paselem: As mentioned above, I think we need to pass a parameter back in here.

Regarding this API, I think passing in a 'data' object is a poor design. The method only requires the jobId, and should not need to know about the structure of some random data structure. If the data structure ever changes, then this method would need to get updated with it, which doesn't make sense. The interface should be kept as clean and simple as possible.

paselem:
This is a pretty significant change, and I did not see any unit tests for it. I would be wary about letting this go into master without having unit tests come along with it.

Also, our naming schema for feature branches follows the pattern feature/<featureName>. Please follow that convention in the future. It helps us identify what kind of work is within the branch.

app/components/market/submit/submit-market-application.component.ts
timotheeguerin: this is actually would not be very useful because we do object.defaultValue after to access it.

app/components/market/submit/submit-market-application.component.ts
t-anjam: Was thinking about corner cases if we decided to add special inputs for the different modes, then which category would they fall into? job or pool? So I tried to kept it general by keeping them all in one form object.

app/components/market/submit/submit-market-application.component.ts
t-anjam: Didn't name this method, but it's because it returns both job and pool template.

app/components/market/submit/submit-market-application.component.ts
t-anjam: will do

app/components/market/submit/submit-market-application.component.ts
t-anjam: It calls the NCJ controller function to submit the job

app/components/market/submit/submit-market-application.component.ts
t-anjam: Will re-add soon.

app/components/market/submit/submit-market-application.component.ts
t-anjam: depends on how we implement the button UX. Right now I added a simple one as a placeholder.

app/components/market/submit/submit-market-application.component.ts
t-anjam: will do

app/components/market/submit/submit-market-application.component.ts
t-anjam: No, there is a default case in the UX to render it as text if it is an unknown type.

app/components/market/submit/submit-market-application.component.ts
t-anjam: If there isn't an advancedType available then it will use the regular type which can be an int.

python/controllers/ncj_controller.py
t-anjam: will do

app/components/market/submit/submit-market-application.component.ts
t-anjam: will do

app/components/market/submit/submit-market-application.component.ts
t-anjam: I agree, I didn't create this function. I removed the data parameter for now till I know where to find the jobID/poolID for all three modes.

app/components/market/submit/submit-market-application.component.ts
timotheeguerin: all inputs should belong to one or the other. and in the case it doesn't we can add another root param

app/components/market/submit/submit-market-application.component.ts
timotheeguerin: well not really needs to be a constant but there should be an intermediary service that does the call. Like we do for the file group service

app/components/market/submit/submit-market-application.component.ts
timotheeguerin: can you then put the button back how it was, and in the other PR you can change it to be part of the summary

app/components/market/submit/submit-market-application.html
t-anjam: agree, will change

app/components/market/submit/submit-market-application.html
```diff
@@ -1,37 +1,223 @@
...
-        <bl-form-page main-form-page [title]="title" [formGroup]="form">
-            <bl-form-section title="Job" subtitle="Job parameters" formGroupName="job">
-                <md-input-container class="form-element" *ngIf="jobNameParam">
-                    <input mdInput type="text" name="name" formControlName="jobName" placeholder="Job name">
-                </md-input-container>
-                <div *ngFor="let param of otherParameters" [ngSwitch]="param.type">
-                    <div *ngSwitchCase="NcjParameterExtendedType.fileGroup">
-                        <bl-file-group-picker [formControlName]="param.id" [label]="param.name" class="form-element" [hint]="param.description">
-                        </bl-file-group-picker>
+    <div *ngIf="mode_state == 'None'">
```
timotheeguerin: `===`

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: will do

python/controllers/ncj_controller.py
t-anjam: done

app/components/market/submit/submit-market-application.html
t-anjam: done

app/components/market/submit/submit-market-application.html
t-anjam: done

python/controllers/ncj_controller.py
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: changed to `if (this.jobTemplate.parameters[key].defaultValue) `

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: done

app/components/market/submit/parameter-input.component.ts
```diff
@@ -0,0 +1,27 @@
+import { Component, Input } from "@angular/core";
+import { NcjJobTemplate, NcjParameter, NcjPoolTemplate } from "app/models";
+import "./parameter-input.scss";
+
+@Component({
+    selector: "bl-parameter-input",
+    templateUrl: "parameter-input.html",
+})
+export class ParameterInputComponent {
+    public static breadcrumb() {
```
timotheeguerin: remove

app/components/market/submit/parameter-input.html
```diff
@@ -0,0 +1,20 @@
+<div (click)="printTypes()">
+    Hello world {{ parameter.id }}
+
+</div>
+
+<div *ngIf="type && parameter" [ngSwitch]="parameter.type">
+    <div *ngSwitchCase="type.string">
+        <input mdInput type="text" name="parameter.name" [placeholder]="parameter.name">
```
timotheeguerin: [formControl]="parameterValue"

app/components/market/submit/parameter-input.component.ts
```diff
@@ -0,0 +1,57 @@
...
+export class ParameterInputComponent implements ControlValueAccessor, OnInit, OnDestroy {
+
+    @Input()
+    public parameter: NcjParameter;
+
+    @Input()
+    public type;
+
+    private _propagateChange: (value: any) => void = null;
+
```
timotheeguerin: public parameterValue = new FormControl();

app/components/market/submit/parameter-input.component.ts
```diff
@@ -0,0 +1,57 @@
...
+    public printTypes(){
+        // do nothing
+    }
+
+    public ngOnInit(): void {
+        // do nothing yet
+    }
+
+    public ngOnDestroy(): void {
+        // do nothing
```
timotheeguerin: unsubscribe

app/components/market/submit/parameter-input.component.ts
```diff
@@ -0,0 +1,57 @@
...
+    @Input()
+    public parameter: NcjParameter;
+
+    @Input()
+    public type;
+
+    private _propagateChange: (value: any) => void = null;
+
+    constructor() {
+        console.log("Contructor:", this.parameter, this.type);
```
timotheeguerin:
```
this._sub = this.parameterValue.valueChanges.subscribe((value) => {
  if(propage) {propagate(value)};
});
```

app/components/market/submit/submit-market-application.html
```diff
@@ -23,87 +19,17 @@
...
-                            </md-input-container>
-                        </div>
-                        <div *ngSwitchDefault>
-                            <md-input-container class="form-element">
-                                <input mdInput type="text" name="param.name" [formControlName]="param.id" [placeholder]="param.name">
-                                <md-hint align="end">{{param.description}}</md-hint>
-                            </md-input-container>
-                        </div>
+                    <div *ngFor="let param of poolParametersWrapper">
+                        <bl-parameter-input [formControlName]="param.id" [parameter]="param" [type]="types" [parameterValues]="poolParams.value"> </bl-parameter-input>
```
timotheeguerin: Don't pass this

app/components/market/submit/parameter-input.component.ts
```diff
@@ -39,12 +38,13 @@ export class ParameterInputComponent implements ControlValueAccessor, OnInit, On
     }
 
     public ngOnInit(): void {
-        this.parameterValue.setValue(this.parameter.defaultValue);
+        // do not need
```
timotheeguerin: remove this one if not using

test/app/components/market/submit/parameter-input.component.spec.ts
```diff
@@ -1,11 +1,23 @@
...
+import { DialogService } from "app/components/base/dialogs";
+import { FileGroupPickerComponent } from "app/components/data/shared";
+import { CloudFilePickerComponent } from "app/components/data/shared/cloud-file-picker";
+import { NcjParameterExtendedType, NcjParameterWrapper } from "app/components/market/submit";
 import { ParameterInputComponent } from "app/components/market/submit";
-import { NcjParameterWrapper } from "app/components/market/submit";
-import { NcjParameterRawType } from "app/models";
+import { BatchApplication, NcjParameterRawType } from "app/models";
+import { StorageService } from "app/services";
+import { Subject } from "rxjs";
```
timotheeguerin: move the library import in the first group above. i.e `@angular/...` `rxjs` ...

test/app/components/market/submit/parameter-input.component.spec.ts
```diff
@@ -40,21 +51,234 @@ fdescribe("ParameterInputComponent", () => {
...
 
-    describe("1st tests", () => {
-        it("true is true", () => expect(true).toBe(true));
+    it("should propogate form control value to class", () => {
+        expect(testComponent.paramControl.value).toBe(component.parameterValue.value);
+    });
+
+    describe("text parameter type", () => {
+        let inputEl: DebugElement;
+        const initialInput: string = "initial input value";
```
timotheeguerin: don't define the type here is is a bit too verbose. Only do the type if it is not obious or on a class attribute

test/app/components/market/submit/parameter-input.component.spec.ts
```diff
@@ -40,21 +51,234 @@ fdescribe("ParameterInputComponent", () => {
...
+
+        it("should show updated input", () => {
+            testComponent.paramControl.setValue(newInput);
+            fixture.detectChanges();
+            fileGroupComponent = fileGroupEl.componentInstance;
+            expect(fileGroupComponent.value.value).toBe(newInput);
+        });
+    });
+
+    describe("fileinput parameter type", () => {
```
timotheeguerin: for this type you alos want to check the containerId given match what's in the formValues

app/components/market/submit/parameter-input.html
```diff
@@ -0,0 +1,35 @@
+<div [ngSwitch]="parameter.type">
+    <div *ngSwitchCase="NcjParameterExtendedType.string">
+        <md-input-container class="form-element">
+            <input mdInput type="text" name="parameter.name" [formControl]="parameterValue" [placeholder]="parameter.name">
```
paselem: Is the 'placeholder' value shown in the UI? If so, should it be a friendly name such as "Parameter name"?

app/components/market/submit/parameter-input.html
```diff
@@ -0,0 +1,35 @@
...
+        [label]="parameter.name" class="form-element" [hint]="parameter.description">
+        </bl-cloud-file-picker>
+    </div>
+    <div *ngSwitchCase="NcjParameterExtendedType.dropDown">
+        <md-select [formControl]="parameterValue" [placeholder]="parameter.name">
+            <md-hint align="end"> {{ parameter.description }} </md-hint>
+            <md-option *ngFor="let opt of parameter.allowedValues" [value]="opt"> {{ opt }} </md-option>
+        </md-select>
+    </div>
+    <div *ngSwitchCase="NcjParameterExtendedType.int">
```
paselem: Should we add a specific type for 'bool'?

app/components/market/submit/market-application.model.ts
```diff
@@ -0,0 +1,83 @@
+import { NcjParameter } from "app/models";
+import { ObjectUtils, log } from "app/utils";
+import * as inflection from "inflection";
+
+export enum Modes {
+    None,
+    NewPoolAndJob,
+    OldPoolAndJob,
```
paselem: Nitpick: "ExistingPoolAndJob" instead of "OldPoolAndJob". Thoughts?

app/components/market/submit/submit-market-application.component.ts
```diff
@@ -75,134 +17,161 @@ export class SubmitMarketApplicationComponent implements OnInit {
...
+    private _parseParameters() {
+        const jobParameters = this.jobTemplate.parameters;
+        const jobTempWrapper: any[] = [];
+        for (let name of Object.keys(jobParameters)) {
+            const param = jobParameters[name];
+            jobTempWrapper.push(new NcjParameterWrapper(name, param));
+        }
+        this.jobParametersWrapper = jobTempWrapper;
+        const poolParameters = this.poolTemplate.parameters;
+        const poolTempWrapper: any[] = [];
```
paselem: shouldn't this be 'let' instead of 'const' if you are going to append to it later?

test/app/components/market/submit/market-application.model.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+    it("should detect int type", () => {
+        parameter = new NcjParameterWrapper("frameEnd", {
+            type: NcjParameterRawType.int,
+            metadata : {
+                description: "description",
+            },
+            defaultValue: 4,
+        });
+        expect(parameter.type).toBe(NcjParameterExtendedType.int);
+    });
```
paselem: May want to add a unit test here to check for the value as well (since this is an int which is a different case than what you have above for string).

paselem: Same goes for cases below.

test/app/components/market/submit/parameter-input.component.spec.ts
t-anjam: removed types for simple variables

test/app/components/market/submit/parameter-input.component.spec.ts
t-anjam: done

app/components/market/submit/parameter-input.html
t-anjam: I think bool scenario can be covered under dropdown case with two options: ['true', 'false'].

app/components/market/submit/market-application.model.ts
t-anjam: done

app/components/market/submit/submit-market-application.component.ts
t-anjam: https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75

image

Since the array variable is not being reassigned then it's safe to use const. Even though its contents are changing.

app/components/market/submit/parameter-input.html
timotheeguerin: remove the name= attribute here

test/app/components/market/submit/market-application.model.spec.ts
t-anjam: done

app/components/market/submit/parameter-input.html
t-anjam: done

app/components/market/submit/parameter-input.component.ts
```diff
@@ -0,0 +1,58 @@
...
+            .subscribe((query: string) => {
+                if (this._propagateChange) {
+                    this._propagateChange(query);
+                }
+            }),
+        );
+    }
+
+    public getContainerFromFileGroup(fileGroup: string) {
+        return fileGroup && `fgrp-${fileGroup}`;
```
ascobie: This is in StorageService.ncjFileGroupPrefix

We could move it into Constants if you like. We shouldn't hard code this anywhere.

PR #632, opened 2017-08-18
ignore errors that we expect to cut down on console noise

### LINKED ISSUE #631: Clean up excessive console errors when task logs are not available on node
The file loader was logging all errors which generates a lot of nose in the console for the Log Output tab then looking at jobs and tasks where either the node no longer exists or the files have been cleaned from the node.

console-noise

ascobie:
Fix #631

Added the FileService.fileIgnoredErrors into the FileLoader config so we can ignore any expected errors and not have them logged. 

Am hoping this hasn't degraded anything else in the app but the console looks great now that its not logging loads of errors.


app/components/task/details/output/task-log.component.ts
```diff
@@ -161,11 +161,21 @@ export class TaskLogComponent implements OnInit, OnChanges, OnDestroy {
...
-                        this.fileSizes[filename] = prettyBytes(props && props.contentLength);
-                    }
+                fileLoader.getProperties().subscribe({
+                    next: (file: File) => {
+                        if (file) {
+                            const props = file.properties;
+                            this.fileSizes[filename] = prettyBytes(props && props.contentLength);
+                        }
+                    },
+                    error: (error) => {
```
timotheeguerin: Because it just ignore the log error it doesn't actually ignore the error.

If you want to ignore the error you'll have to do .catch but I don't think that's a good idea

timotheeguerin: Here i should just have error: () => null

PR #659, opened 2017-09-01
Application package picker for pool creation

### LINKED ISSUE #658: Add application package selector to pool creation
Need to be able to create pools with application packages

ascobie:
Fix #658

Create Pool
pool

Clone or Add Task
task


app/styles/vendor/material-theme.scss
```diff
@@ -128,6 +128,10 @@ md-dialog-container.mat-dialog-container {
     width: auto !important;
     height: auto !important;
     max-height: 80vh;
+
+    form {
```
timotheeguerin: hhm don't like this, why on all forms?

app/components/pool/base/app-packages/app-package-picker.html
```diff
@@ -0,0 +1,4 @@
+<bl-editable-table [formControl]="files">
```
timotheeguerin: files?

app/styles/vendor/material-theme.scss
ascobie: its not on all forms, only on the "mat-dialog-container" as the padding that was on here has been removed and the dialog's look funny with no padding. ill add a before and after pic.

app/components/pool/base/app-packages/app-package-picker.html
ascobie: yeah, not done yet. used the resource file picker as a base.

app/styles/vendor/material-theme.scss
ascobie: before 

no-padding

after

padding

timotheeguerin: I think this was because I try to remove some padding with the simple form but I reverted that in the upload pr

timotheeguerin:
can you post screenshot of this?

PR #677, opened 2017-09-08
Internal: find an available port to connect the python rpc server

### LINKED ISSUE #676: Find an open port for the python server to connect to
Currently the port is hard coded which might conflict with other apps and/or if the rpc server wasn't killed correctly because of a crash

timotheeguerin:
fix #676

### DIFF
--- a/src/client/client-constants.ts
+++ b/src/client/client-constants.ts
@@ -1,9 +1,27 @@
+import * as net from "net";
 import * as path from "path";
-
 /**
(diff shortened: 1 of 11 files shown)


src/client/client-constants.ts
```diff
@@ -26,10 +44,17 @@ const urls = {
 
 const isAsar = process.mainModule.filename.indexOf("app.asar") !== -1;
 
+const pythonServerPort = {
+    // dev: Promise.resolve(8765),
```
timotheeguerin: change

PR #681, opened 2017-09-08
Feature/ncj remove validator

t-anjam:
Remove required validator when a defaultValue is present and it's value is an empty string. Needed if we want to give users optional arguments. Might want to consider adding a "required" field in the job/pool template for parameters in the future.

### DIFF
--- a/app/components/market/submit/submit-market-application.component.ts
+++ b/app/components/market/submit/submit-market-application.component.ts
@@ -132,10 +132,15 @@ export class SubmitMarketApplicationComponent implements OnInit {
         const templateFormGroup = {};
         for (let key of templateParameters) {
             let defaultValue = null;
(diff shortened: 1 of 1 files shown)


app/components/market/submit/submit-market-application.component.ts
```diff
@@ -132,10 +132,14 @@ export class SubmitMarketApplicationComponent implements OnInit {
         const templateFormGroup = {};
         for (let key of templateParameters) {
             let defaultValue = null;
-            if (template.parameters[key].defaultValue) {
+            let validator = Validators.required;
+            if (template.parameters[key].defaultValue || template.parameters[key].defaultValue === "") {
```
timotheeguerin: exists(template.parameters[key].defaultValue)

PR #706, opened 2017-09-19
Merge latest master into NCJ

ascobie:
App package support for pools and tasks
fix absolute path bug in file explorer
fix background task label


timotheeguerin:
yeah, watch out when you merge this one as you don't want to squash merge or rebase merge. I would just do the merge manually and push or renable merge commit then merge and disable it again

PR #721, opened 2017-09-22
grammar issues and fix blob file explorer

ascobie:
change a couple of words in notifications
fix the blob file explorer by removing training slash that was messing up the storage prefix.

### DIFF
--- a/app/components/task/details/output/task-outputs.component.ts
+++ b/app/components/task/details/output/task-outputs.component.ts
@@ -61,7 +61,9 @@ export class TaskOutputsComponent implements OnChanges, OnDestroy {
     }
 
     public ngOnDestroy() {
-        this.workspace.dispose();
+        if (this.workspace) {
+            this.workspace.dispose();
+        }
     }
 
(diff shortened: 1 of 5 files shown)


app/components/task/details/output/task-outputs.component.ts
```diff
@@ -85,7 +85,7 @@ export class TaskOutputsComponent implements OnChanges, OnDestroy {
             });
             nodeNavigator.init();
 
-            const taskOutputPrefix = `${this.task.id}/`;
+            const taskOutputPrefix = `${this.task.id}`;
```
timotheeguerin: I think we should make the command support having a basePath ending or not with / instead of joining / and then breaking that.

ascobie: agree though i don't think there is any need to have a / here either. given that at the moment the prefix is only used by blob storage queries and the prefix has never ended in a / before. So i suggest we remove it here and also make the command handle having a path segment ending in a /

PR #744, opened 2017-10-06
Delete folder or file from file group

### LINKED ISSUE #733: Allow delete folder/file from the file group context menu.
tree view can delete a folder or a file: 

delete

Needs to be used sparingly on folders as it deletes files one at a time as per storage API constraints.

ascobie:
Fix #733

Only allow deletion from a storage container. Ignore files and folders on compute nodes.
Storage API only allows deleting blobs one at a time. 

delete


app/components/file/browse/blob-files-browser/blob-files-browser.component.ts
```diff
@@ -36,6 +40,7 @@ export class BlobFilesBrowserComponent implements OnChanges, OnDestroy {
         if (inputs.upload) {
             this.fileExplorerConfig = {
                 canDropExternalFiles: Boolean(this.upload),
+                canDeleteFiles: true,
```
timotheeguerin: same as above `Boolean(this.delete)`

app/components/file/browse/blob-files-browser/blob-files-browser.component.ts
```diff
@@ -52,11 +57,40 @@ export class BlobFilesBrowserComponent implements OnChanges, OnDestroy {
...
     }
 
+    public handleDeleteEvent(event: FileDeleteEvent) {
+        const { path } = event;
+        console.log("BlobFilesBrowserComponent.handleDeleteEvent: ", event);
+
+        // TODO: Check with Tim as to whether i can chain 2 background tasks to run one after another.
+        // In which case most of this can be moved into DataContainerFilesComponent.
+        // task 1: get files to delete
+        // task 2: delete files
```
timotheeguerin: you could but I don't think thats a good idea, it would look weird. Maybe what you could do is have it send back an observable that resolve files instead of the actual files so the waiting can be done in the background task

timotheeguerin: but otherwise I think its all good like that for now until we make that generic FileSystem

app/components/file/browse/blob-files-browser/blob-files-browser.component.ts
```diff
@@ -52,11 +57,40 @@ export class BlobFilesBrowserComponent implements OnChanges, OnDestroy {
...
+
+        this.dialogService.confirm(`Delete files`, {
+            description: event.isDirectory
+                ? `All files will be deleted from the folder: ${path}`
+                : `The file '${FileUrlUtils.getFileName(path)}' will be deleted.`,
+            yes: () => {
+                const listParams = { recursive: true, startswith: path };
+                const data = this.storageService.listBlobs(Promise.resolve(this.container), listParams);
+                return data.fetchAll().flatMap(() => data.items.take(1)).map((items) => {
```
timotheeguerin: i would not use .map here because it won't run unless you subscribe to it. I know this funciton is being passed up to the dialog which subscribe to it but it is a bit confusing to actually do state actions in the map. I would just have it 
const obs = ...
obs.subscribe(...)
return obs;

app/components/data/details/data-container-files.component.ts
```diff
@@ -64,4 +63,39 @@ export class DataContainerFilesComponent implements OnDestroy {
...
+        const taskTitle = `Delete ${fileCount} files from ${this.container.name}`;
+        console.log(`handleFileDelete with ${fileCount} files`);
+
+        // NOTE: Just pretending to delete the files at the moment as i need to check with the
+        // storage team as to the best way to batch delete a heap of files as there could be lots
+        // of them in a container.
+
+        return this.backgroundTaskService.startTask(taskTitle, (task) => {
+            let deleted = 0;
+            let observable = Observable.interval(250).take(fileCount);
```
timotheeguerin: const

app/components/data/details/data-container-files.component.ts
```diff
@@ -64,4 +63,47 @@ export class DataContainerFilesComponent implements OnDestroy {
...
+            observable.subscribe({
+                next: (i) => {
+                    deleted++;
+                    return this.storageService.deleteBlobIfExists(this.container.id, files[i].name).subscribe({
+                        next: (response) => {
+                            task.name.next(`${taskTitle} (${deleted}/${fileCount})`);
+                            task.progress.next(deleted / fileCount * 100);
+
+                            if (response && blobCache) {
+                                blobCache.deleteItem(files[i]);
```
timotheeguerin: shouldn't it be the deleteBlobIfexists that remove from the cache. I feel this would be better. THat way you don't have to do that everytime

app/components/data/details/data-container-files.component.ts
```diff
@@ -64,4 +63,47 @@ export class DataContainerFilesComponent implements OnDestroy {
...
+                        },
+                        error: (error) => {
+                            log.error("Failed to delete blob", error);
+                        },
+                    });
+                },
+                complete: () => {
+                    task.progress.next(100);
+                    // tslint:disable-next-line:max-line-length
+                    const message = `${deleted} files were successfully removed from the file group: ${this.container.name}`;
```
timotheeguerin: just a note this is where I think starting to think about i18n would be good. Long string like that don't look good in the code

app/components/data/details/data-container-files.component.ts
```diff
@@ -64,4 +63,47 @@ export class DataContainerFilesComponent implements OnDestroy {
...
+                    this.notificationService.success("Removed files from group", message, { persist: true });
```
timotheeguerin: do we really want to persist that one?

app/components/file/browse/blob-files-browser/blob-files-browser.component.ts
```diff
@@ -52,11 +57,36 @@ export class BlobFilesBrowserComponent implements OnChanges, OnDestroy {
...
+
                 return Observable.of(null);
             },
         });
     }
 
+    public handleDeleteEvent(event: FileDeleteEvent) {
+        const { path } = event;
+        this.dialogService.confirm(`Delete files`, {
+            description: event.isDirectory
```
timotheeguerin: can you make a description variable no to have this ternary condition inline

app/components/file/browse/blob-files-browser/blob-files-browser.component.ts
```diff
@@ -52,11 +57,36 @@ export class BlobFilesBrowserComponent implements OnChanges, OnDestroy {
...
+                ? `All files will be deleted from the folder: ${path}`
+                : `The file '${FileUrlUtils.getFileName(path)}' will be deleted.`,
+            yes: () => {
+                const listParams = { recursive: true, startswith: path };
+                const data = this.storageService.listBlobs(this.container, listParams);
+                const obs = data.fetchAll().flatMap(() => data.items.take(1));
```
timotheeguerin: add .shareReplay(1) at the end so it doesn't trigger the flatMap for every subscription

app/services/file/file-navigator/file-navigator.ts
```diff
@@ -63,13 +71,19 @@ export class FileNavigator {
...
+        this._cache = config.cache;
     }
 
     /**
      * Load the inital data
      */
     public init() {
         this._loadFilesInPath("");
+        if (this._cache) {
+            this._fileDeleted = this._cache.deleted.subscribe((key: string) => {
```
timotheeguerin: nitpick:`_fileDeletedSub`

app/components/data/details/data-container-files.component.ts
ascobie: on a .subscribe(() => { /* remove from cache */ })?

probably a very good idea and one that i am not sure why i didn't think of :)

app/components/data/details/data-container-files.component.ts
ascobie: for sure, and every time i see loads of UI text in the code i think the same thing.

app/components/data/details/data-container-files.component.ts
ascobie: probably not, though i saw it was in a few other places. can remove.

timotheeguerin: yeah error and warning are persisted by default but info and success disapear.
I force it to displayed for stuff where you have actions that you might want to do

app/components/data/details/data-container-files.component.ts
```diff
@@ -66,44 +66,19 @@ export class DataContainerFilesComponent implements OnDestroy {
...
-                complete: () => {
-                    task.progress.next(100);
-                    // tslint:disable-next-line:max-line-length
-                    const message = `${deleted} files were successfully removed from the file group: ${this.container.name}`;
-                    this.notificationService.success("Removed files from group", message, { persist: true });
-                },
-            });
-
-            return observable;
+        // TODO: would like to get the deleted count from this call.
```
timotheeguerin: use next instead of complete and make the service return the number of files

app/components/data/details/data-container-files.component.ts
```diff
@@ -66,44 +66,19 @@ export class DataContainerFilesComponent implements OnDestroy {
...
+        this.storageService.deleteFilesFromContainer(this.container, files).subscribe({
+            complete: () => {
+                // tslint:disable-next-line:max-line-length
+                const message = `The files were successfully removed from the file group: ${this.container.name}`;
+                this.notificationService.success("Removed files from group", message);
+            },
+            error: (error) => {
+                // tslint:disable-next-line:max-line-length
+                const message = `Not all files were able to be removed from the file group: ${this.container.name}`;
+                this.notificationService.error("Error while deleting files", message, { persist: true });
```
timotheeguerin: don't need to specify presist true for error its the default

app/services/storage.service.ts
```diff
@@ -199,7 +203,51 @@ export class StorageService {
...
+        const fileCount = files.length;
+        const taskTitle = `Delete ${fileCount} files from ${container.name}`;
+        let deleted = 0;
+
+        return this.backgroundTaskService.startTask(taskTitle, (task) => {
+            // NOTE: slight pause in-between deletes to ease load on storage account
+            const observable = Observable.interval(100).take(fileCount);
+            observable.subscribe({
+                next: (i) => {
+                    deleted++;
```
timotheeguerin: isn't deleted the same as i?

app/components/data/details/data-container-files.component.ts
timotheeguerin: do you mean the number of files(because you can just do files.length above) or the number of files that actually existsed and were deleted?

app/services/storage.service.ts
ascobie: yes, though i need to change it to be the actual number of deleted files.

timotheeguerin: but actually what do you expect to show if the number is different? Some files wasn't deleted successfully or some files were already deleted?

ascobie: i expected i would show a warning that not all files were able to be deleted. but maybe i just shouldn't bother ....

PR #818, opened 2017-10-31
Feature/debug aad auth

Sarkarmuhammed:


src/client/authentication/authentication-window.ts
```diff
@@ -13,11 +13,23 @@ export class AuthenticationWindow extends UniqueWindow {
         });
 
         // Uncomment to debug auth errors
-        // this._window.webContents.openDevTools();
+        window.webContents.openDevTools();
```
ascobie: comment this out as above.

src/client/main.ts
```diff
@@ -20,7 +20,7 @@ function startApplication() {
     });
 
     // Uncomment to view why windows don't show up.
-    // batchLabsApp.debugCrash();
+    batchLabsApp.debugCrash();
```
ascobie: comment this out as above.

PR #821, opened 2017-11-01
export ncj recent template paramters

ascobie:
Add ability to export template parameters: 

download

Note: Dave would also like to show the associated CLI command for running the template which i will add in the submit template UI. That will be another PR.

Add padding to local templates heading:

local-template

### DIFF
--- a/app/components/market/home/recent-template-list/recent-template-list.component.ts
+++ b/app/components/market/home/recent-template-list/recent-template-list.component.ts
@@ -1,7 +1,8 @@
 import { Component, OnDestroy } from "@angular/core";
+import { remote } from "electron";
 import { Subscription } from "rxjs";
 
-import { Modes } from "app/components/market/submit";
+import { NcjTemplateMode } from "app/models";
 import { NcjTemplateService, RecentSubmission } from "app/services";
 import "./recent-template-list.scss";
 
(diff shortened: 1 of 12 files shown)


app/components/market/home/recent-template-list/recent-template-list.component.ts
```diff
@@ -25,14 +26,36 @@ export class RecentTemplateListComponent implements OnDestroy {
...
+        switch (mode) {
+            case NcjTemplateMode.NewPool:
+                return "Pool for later use";
+            case NcjTemplateMode.ExistingPoolAndJob:
+                return "Job with existing pool";
+            case NcjTemplateMode.NewPoolAndJob:
+                return "Job with auto-pool";
+        }
+    }
+
```
ascobie: I changed these to be more inline with the action that the user took when running the template. Hope you don't mind.

app/components/market/local-template-browser/local-template-browser.scss
```diff
@@ -9,7 +9,7 @@ bl-local-template-browser {
 
     > .header {
         height: 40px;
-        padding: 0 20px;
+        padding: 10px 20px;
     }
```
ascobie: padding for the local template heading between the breadcrumb and then heading to bring it in line with the other NCJ screens.

ascobie:
going to commit this. it's not the ideal solution as we discussed in the meeting this morning, but its a valid start and can be modified to include the other required features.

PR #825, opened 2017-11-02
Click link to show readme file from action chooser

ascobie:
Opens the MD file in Chrome via GitHub. 
Only 3ds Max and ImageMagick have full readme files at the moment. Will work on the rest tomorrow.

capture

### DIFF
--- a/app/components/market/application-action/choose-action.html
+++ b/app/components/market/application-action/choose-action.html
@@ -4,6 +4,7 @@
         <h1>
             <img class="logo" [src]="application.icon" height="30" width="30" alt="logo" *ngIf="application?.icon">
             <span>Action selection for <span [title]="application.description">{{application?.name}}</span></span>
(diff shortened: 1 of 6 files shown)


app/components/market/application-action/choose-action.html
```diff
@@ -4,6 +4,7 @@
         <h1>
             <img class="logo" [src]="application.icon" height="30" width="30" alt="logo" *ngIf="application?.icon">
             <span>Action selection for <span [title]="application.description">{{application?.name}}</span></span>
+            <span>(<a (click)="openLink(application?.readme)" href="javascript:void(0)">view readme</a>)</span>
```
paselem: Should we wrap this in an 'if' block in case the readme doesn't exist?

ascobie: Like the Icon file we just "assume" the readme file is there. If this becomes an issue, or we end up needing more than one readme (per folder/action), then we can look at a more encompassing solution by adding the icon and readme settings into the index.json file.

PR #832, opened 2017-11-06
Batch Application context menu and delete refactor.

### LINKED ISSUE #776: Add context menu to app packages quick-list
This looks to be the only remaining list that has no context menu

### LINKED ISSUE #831: Auto delete package versions when deleting an application package
Currently you cannot delete an application (app package), if it has associated package versions. This means you have to delete the versions manually before deleting the parent.

Change the delete action to list and auto delete the versions before deleting the parent app package.

ascobie:
Fix #776
Fix #831

- [x] Add context menu to batch application quick-list
- [x] Pass in enabled state to bl-delete-button
- [x] Delete action refactor to delete versions first
- [x] Disable delete multiple on quick-list if a selected application has allow updates to be false.


ascobie:
disabled quick-list delete as we cannot delete an application that have allowUpdates set to false and there is no easy way to check this and disable the quick-list delete.

Can be re-enabled in the future when i have time to look into a better solution, but can't see people wanting to bulk delete applications as we are limited to 20 anyway.

PR #859, opened 2017-11-13
Internal: Refactor RxListProxy into new system

### LINKED ISSUE #814: Refactor rx-list-proxy to a new system
From #795 2nd part. Replacing the rx-list-proxy to have better and cleaner use.

list() /=> Return Observable of List of objects(Can take a continuation token)
listAll() /=> Just an helper method which calls list until no more items and return the result. (Maybe could have some progress tracker?)

listView() /=> returns a `ListView` (Same interface as current `RxListProxy`)

timotheeguerin:
fix #814


app/services/file-service.ts
```diff
@@ -79,6 +81,36 @@ export class FileService extends ServiceBase {
             getFn: (client, { poolId, nodeId, filename }) =>
                 client.file.getComputeNodeFileProperties(poolId, nodeId, filename),
         });
+
+        this._taskFileListGetter = new BatchListGetter(File, this.batchService, {
+            cache: (params) => this.getTaskFileCache(params),
+            list: (client, { jobId, taskId }, options) => {
+                const batchOptions = { ...options };
+                console.log("Options", options);
```
ascobie: remove

PR #863, opened 2017-11-14
App crashes when a pool has no start task

ascobie:


app/components/pool/details/pool-configuration.html
```diff
@@ -54,7 +54,7 @@
         <bl-property-group label="Start task settings" [collapsed]="true" class="wide" [edit]="editStartTask">
             <bl-text-property label="Command line" [value]="startTask.commandLine" [wrap]="true"></bl-text-property>
             <bl-text-property label="Max retry count" [value]="startTask.maxTaskRetryCount"></bl-text-property>
-            <bl-text-property label="User identity" [value]="decorator.startTask.userIdentitySummary"></bl-text-property>
+            <bl-text-property label="User identity" [value]="decorator.startTask?.userIdentitySummary"></bl-text-property>
```
timotheeguerin: shouldn't we actually not even show properties here if start task is null?

ascobie: sorry missed this comment. we could probably hide the properties, but not the entire section as that allows for editing/adding the start task.

PR #891, opened 2017-11-21
Feature: Create an empty file group

### LINKED ISSUE #826: Allow users to create empty file groups
This is required for running NCJ jobs as you generally don't want to re-use the input file group containers for your job outputs. We need to be able to create empty file groups for job outputs.

TODO: check with Anna that this can be done.

timotheeguerin:
fix #826


app/components/data/home/data-home.component.ts
```diff
@@ -57,4 +72,8 @@ export class DataHomeComponent implements OnDestroy {
     private _updateFilter() {
         this.filter = this.quickFilter;
     }
+
+    private _createEmptyFileGroup(name: string) {
+        return this.storageService.createContainer(name);
+    }
 }
```
ascobie: does this prefix the container with fgrp-?

app/services/storage/blob-storage-client-proxy.ts
```diff
@@ -275,7 +275,29 @@ export class BlobStorageClientProxy {
...
+                }
+            });
+        });
+    }
+
+    /**
+     * Marks the specified container for deletion. The container and any blobs contained within
+     * it are later deleted during garbage collection.
+     * http://azure.github.io/azure-storage-node/BlobService.html#deleteContainer__anchor
+     *
```
ascobie: comments don't match the function

app/components/data/home/data-home.component.ts
timotheeguerin: yeah, as storage is crashing right now couldn't check it was working

PR #900, opened 2017-11-22
Internal: Use route guard to prevent navigation until account is selected

johnnyzhang82:
Implement a guard and directive that hide navigation if users have no account selected.


app/components/base/guards/navigation-guard.component.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from "@angular/core";
+import { ActivatedRouteSnapshot, CanActivate, RouterStateSnapshot } from "@angular/router";
+import { AccountService } from "app/services";
+
+@Injectable()
+export class NavigationGuard implements CanActivate  {
+    constructor(
+        // private router: Router,
```
timotheeguerin: remove?

app/components/base/guards/navigation-guard.component.ts
```diff
@@ -0,0 +1,18 @@
...
+        private accountService: AccountService,
+    ) {
+    }
+
+    public canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
+        return this.accountService.currentAccountId.map((accountId) => {
```
timotheeguerin: change to Boolean instead of !!

PR #950, opened 2017-12-04
Feature: Always include active entity in the list when looking at its details

### LINKED ISSUE #199: Goto directly to an entity doesn't show the entity in the quicklist
Clicking on the pool from the job props will load the pool but if you have too many pools it won't show it on the quick list on the left.

We could add the pool to the top of the list if not already there like when adding a new pool

timotheeguerin:
fix  #199


app/components/pool/browse/pool-list.component.ts
```diff
@@ -153,6 +156,10 @@ export class PoolListComponent extends ListOrTableBase implements OnInit, OnDest
         return pool.id;
     }
 
+    public activePoolChanged(data: any) {
+        console.log("Active?/", data);
```
timotheeguerin: remove

app/components/pool/browse/pool-list.component.ts
```diff
@@ -50,7 +51,7 @@ export class PoolListComponent extends ListOrTableBase implements OnInit, OnDest
         this._filter = filter;
 
         if (filter.isEmpty()) {
-            this.data.setOptions({});
+            this.data.setOptions({ pageSize: 2 });
```
timotheeguerin: remove

app/models/navigable-record.ts
```diff
@@ -0,0 +1,21 @@
+export enum PinnedEntityType {
+    Application,
+    Job,
+    Pool,
+    FileGroup,
+}
+
```
ascobie: weren't you removing the PinnedEntityType ?

PR #952, opened 2017-12-05
Feature: use ratecard api for pricing

### LINKED ISSUE #684: Pool estimated cost take rendering license into account
For now the estimated cost only shows the price of the nodes.

Need an api to return the resource id for it.

### LINKED ISSUE #857: Pricing is broken
The API call to: 

https://s2.billing.ext.azure.com/api/Billing/Subscription/GetSpecsCosts?api-version=2016-09-01

Is failing with: 

**_401 - Unauthorized: Access is denied due to invalid credentials.
You do not have permission to view this directory or page using the credentials that you supplied._**

timotheeguerin:
fix #857 (Pricing broken)
fix #684 (Rendering license price taken into account)
license-pricing


ascobie:
will this work for getting the render licenses as well? we should have the meter id's for them.

PR #959, opened 2017-12-07
Added error tooltip and banner for invalid account

### LINKED ISSUE #890: Should show better message to user when account 404's
While scanning through the account quicklist, if an account has been deleted since last time you loaded the application, it will 404. Without looking at the logs/console it's hard to see why the account didn't load. 

We should probably show a better message box to the user in the case of an account 404. The account selector in the title bar does show a red triangle, but that is not easy to see with the red on dark blue, so any users with color blindness issues may not notice it.

capture

johnnyzhang82:
Fixed #890, added native tooltip for invalid account error message when hovering on invalid account.
Tooltip:
batchlab

Error Banner (Code | Message format):
batchlaberror


ascobie:
With this one, i think that the tooltip is a good start, but i think we need to go further. 

I was thinking an error message banner like this would be a good idea as the tooltip is still a little obscure. 

capture

PR #962, opened 2017-12-07
Batch account favorite list item context menu implementation, etc

johnnyzhang82:
- added context menu (select, remove) in favorite batch account list
- fixed an issue that batch account error banner not reset
- removed navigation menu no-userinteract implementation since there is already a route guard 

image


app/components/account/browse/account-dropdown.component.ts
```diff
@@ -46,4 +49,28 @@ export class AccountDropDownComponent implements AfterViewInit {
             this.currentAccountInvalidError = error;
         });
     }
+
+    public openContextMenu(account: AccountResource) {
+        const items = [
```
timotheeguerin: if not enabled just return(don't show context menu) instead of having all enabled set to false

PR #969, opened 2017-12-11
Internal: Move aad code from renderer to node context

### LINKED ISSUE #963: Extract AAD logic to be outside of the angular service into the node environment
The problem is that it will try to relogin for every window while you technically don't need to. The login should be handled at the node level that way multiple windows can share the same tokens.

This will make #74 experience much nicer.

timotheeguerin:
fix #963

This also include moving other pieces of code in a new common folder(Modules that can be used in both renderer and node context)


app/services/app-insights/app-insights-api.service.ts
```diff
@@ -5,8 +5,9 @@ import { Observable } from "rxjs";
 
 import { ServerError } from "app/models";
 import { AccountService } from "app/services/account.service";
-import { AccessToken, AdalService } from "app/services/adal";
+import {  AdalService } from "app/services/adal";
```
johnnyzhang82: extra space

PR #993, opened 2017-12-21
Feature: Inbound endpoints support

### LINKED ISSUE #965: Support for inbound endpoints
To setup and use VRay distributed rendering in Batch Labs we'll need inbound endpoint support.  There's two components to this.

1. NCJ support for specifying inbound endpoints on pools.
2. BL support to display the pool and compute node endpoint information.

johnnyzhang82:
fix #965 
-added pool endpoint configuration in create pool form
-fixed a bug in job container setting validators not being reset

nat2
nat3
nat0
nat1


app/components/base/form/editable-table/editable-table-column.component.ts
```diff
@@ -9,5 +9,8 @@ export class EditableTableColumnComponent {
 
     @Input() public type: string = "text";
```
timotheeguerin: Can you make this type to be an enum(with string values)

app/components/base/form/editable-table/editable-table-column.component.ts
```diff
@@ -9,5 +9,8 @@ export class EditableTableColumnComponent {
...
 
+    /** Select options when type is 'select' */
+    @Input() public select: string[];
```
timotheeguerin: I would call that options maybe

app/components/pool/pool-endpoint/inbound-nat-pool-picker.component.ts
```diff
@@ -0,0 +1,122 @@
...
+        this._propagateChange = fn;
+    }
+
+    public registerOnTouched() {
+        // Do nothing
+    }
+
+    public validate(c: FormControl) {
+        const valid = this.form.valid;
+        if (!valid) {
```
timotheeguerin: `if(valid) {return null;}`?

app/components/pool/pool-endpoint/inbound-nat-pool-picker.component.ts
```diff
@@ -0,0 +1,122 @@
...
+        // tslint:disable:no-forward-ref
+        { provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => InboundNATPoolPickerComponent), multi: true },
+        { provide: NG_VALIDATORS, useExisting: forwardRef(() => InboundNATPoolPickerComponent), multi: true },
+    ],
+})
+export class InboundNATPoolPickerComponent implements ControlValueAccessor, Validator, OnDestroy {
+    @Input() public inboundNATPools: InboundNATPool[];
+
+    public InboundEndpointProtocol = InboundEndpointProtocol;
+    public minPort = EndpointHelper.MININUM_PORT;
```
timotheeguerin: can't you just pass the EndpointHelper here instead of all those values, also I feel like we don't really name our constant ALL_CAPS so not sure if we should keep it like that

app/components/base/form/editable-table/editable-table.component.ts
```diff
@@ -109,6 +109,10 @@ export class EditableTableComponent implements ControlValueAccessor, Validator,
         return row;
     }
 
+    public trackIndex(index, row: any) {
```
timotheeguerin: here actually you could try by the item as it's an option of type string. should be more efficient(Though doesn't really matter here as its small list that doesn't get updated)

johnnyzhang82: I will use trackRows which is already defined above.

PR #1002, opened 2018-01-05
Feature: permissions

### LINKED ISSUE #1000: Rbac permission support. Disable action in batchlabs if user doesn't have write permission
Support disabling actions and part of batchlabs if user doesn't have permission.
For now there is just read and write permission on the entire account but we should design it to support other permissions

johnnyzhang82:
Added 'write' permission check for clickable component,  and disable any extended controls if user only has 'reader' permission of current batch account.
 fix #1000

image


app/components/base/buttons/button.component.ts
```diff
@@ -73,6 +77,21 @@ export class ButtonComponent extends ClickableComponent implements OnChanges {
         if ("disabled" in changes) {
             this.tabindex = this.disabled ? "-1" : "0";
         }
+        if ("enforcePermission" in changes) {
+            this._sub = this.authHttpService.isResourceReadOnly().subscribe(resouceReadOnly => {
```
timotheeguerin: you need to unsubsribe of the sub here first if applicable

app/components/base/buttons/button.component.ts
```diff
@@ -73,6 +77,21 @@ export class ButtonComponent extends ClickableComponent implements OnChanges {
...
+                if (resouceReadOnly) {
+                    this.disabled = true;
+                    this.tabindex = "-1";
+                    this.title += " (You don't have permission to perform this action)";
```
timotheeguerin: Wouldn't this keep appending to the title?

app/components/base/buttons/button.component.ts
```diff
@@ -33,13 +34,14 @@ export enum SubmitStatus {
...
 })
-export class ButtonComponent extends ClickableComponent implements OnChanges {
+export class ButtonComponent extends ClickableComponent implements OnChanges, OnDestroy {
     public SubmitStatus = SubmitStatus;
 
     @Input() public action: ButtonAction;
     @Input() public icon: string;
     @Input() public title: string;
     @Input() public tooltipPosition: string = "above";
+    @Input() public enforcePermission?: boolean;
```
timotheeguerin: can we name this more `permission="write"` as it will give us more flexibility into adding other permissions

app/services/authorization-http.service.ts
```diff
@@ -0,0 +1,84 @@
...
+            }
+        }
+        // If user only has 'Reader' role without any 'Write' roles, button should be disabled
+        // Note that user could be assigned to multiple roles at same time (Reader, Owner, Contributor),
+        // in this case, permission should be checked from highest to lowest
+        return !actions.includes(BatchAccountPermission.ReadWrite) && actions.includes(BatchAccountPermission.Read);
+    }
+
+    public recursiveRequest(uri: string, options?: RequestOptionsArgs): Observable<Response> {
+        options = mergeOptions(options, RequestMethod.Get);
```
timotheeguerin: can't you just call armService.get here instead?

app/components/base/buttons/button.component.ts
```diff
@@ -73,6 +78,7 @@ export class ButtonComponent extends ClickableComponent implements OnChanges {
         if ("disabled" in changes) {
             this.tabindex = this.disabled ? "-1" : "0";
         }
+        this.onPermissionChange(changes);
```
timotheeguerin: super.ngOnChanges()

app/services/authorization-http.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    actions: string[];
+    noActions: string[];
+}
+
+export enum BatchAccountPermission {
+    Read = "*/read",
+    ReadWrite = "*",
+}
+
+export type Permission = "none" | "read" | "write";
```
timotheeguerin: enum Permission {
  none: "none",
...
}

app/components/base/buttons/clickable/clickable.component.ts
```diff
@@ -1,18 +1,48 @@
...
+    }
+
+    public onPermissionChange(changes: SimpleChanges): void {
+        this.subtitle = "";
+        if (changes.permission) {
+            this._clearSubscription();
+            if (changes.permission.currentValue === "write") {
+                this._sub = this.authHttpService.getResourcePermission().subscribe((permission: Permission) => {
+                    // when user only got 'read' permission
+                    if (permission === "read") {
```
timotheeguerin: if(this.permission === "write" && permission === "read") {
// disable
} else {
// enable
}

app/components/base/buttons/clickable/clickable.component.ts
```diff
@@ -1,18 +1,48 @@
...
+    public subtitle = "";
+    private _sub: Subscription;
+    public constructor(private authHttpService: AuthorizationHttpService) {
+    }
+
+    public onPermissionChange(changes: SimpleChanges): void {
+        this.subtitle = "";
+        if (changes.permission) {
+            this._clearSubscription();
+            if (changes.permission.currentValue === "write") {
```
timotheeguerin: if(this.permission)

app/components/base/buttons/clickable/clickable.component.ts
```diff
@@ -1,18 +1,58 @@
...
+    public subtitle = "";
+
+    private _sub: Subscription;
+
+    constructor(private authHttpService: AuthorizationHttpService) {
+    }
+
+    public ngOnChanges(changes: SimpleChanges): void {
+        if ("disabled" in changes) {
+            this.tabindex = this.disabled ? "-1" : "0";
```
timotheeguerin: how about you just move this at the end of ngChanges and remove the if statement so it computes it everytime and you can remove the reasignment below

PR #1011, opened 2018-01-13
Feature: Account monitoring charts

### LINKED ISSUE #111: Add charts to account landing page
Use the azure monitoring api to get this graphs. It should have small boxes on the account page then allow you to focus a graph and view details

- [x] Total tasks running, completed, queued as "live feed". (May need to wait for counting API)
- [x] States of nodes
- [x] Number of cores used
- [x] States of nodes

johnnyzhang82:
Fixed #111 Added 4 different monitor charts in batchlab landing page (same with portal overview)
-dedicated/lowpri corecount
-failed task
-task states
-node states

Also allow user to pick different time frame for these charts (1 hour -default, 1 day and 1 week)

image


paselem:
Very very cool! Thanks for contributing this. Minor nitpick: The cards above and below (overview, job status, pool status, etc) seem to have shadows on them, while the charts do not. Please update to be consistent.

app/services/monitor-http/core-count-metrics.ts
```diff
@@ -0,0 +1,11 @@
+import { MonitorChartAggregation, MonitorChartMetrics, MonitorMetricsBase } from "./monitor-metrics-base";
+
+const statesColor = [
+    { state: "corecount", color: "#1C3F95" },
```
ascobie: Should all color values be in CSS rather than hard coded? We are moving towards themes so it would make sense to move all of these values (and all other # values below) into a CSS file.

app/services/monitor-http/task-states-metrics.ts
```diff
@@ -0,0 +1,18 @@
+import { MonitorChartAggregation, MonitorChartMetrics, MonitorMetricsBase } from "./monitor-metrics-base";
+
+const statesColor = [
+    { state: "taskstartevent", color: "#ffa90d" },
+    { state: "taskcompleteevent", color: "#388e3c" },
+];
```
ascobie: CSS

app/services/monitor-http/node-states-metrics.ts
```diff
@@ -0,0 +1,28 @@
+import { NodeState } from "app/models";
+import { MonitorChartAggregation, MonitorChartMetrics, MonitorMetricsBase } from "./monitor-metrics-base";
+
+const statesColor = [
+    { state: NodeState.starting, color: "#1C3F95" },
+    { state: NodeState.idle, color: "#be93d9" },
+    { state: NodeState.running, color: "#388e3c" },
+    { state: NodeState.startTaskFailed, color: "#aa3939" },
+    { state: NodeState.rebooting, color: "#ff755c" },
+];
```
ascobie: CSS

app/services/monitor-http/failed-task-metrics.ts
```diff
@@ -0,0 +1,11 @@
+import { MonitorChartAggregation, MonitorChartMetrics, MonitorMetricsBase } from "./monitor-metrics-base";
+
+const statesColor = [
+    { state: "taskfailevent", color: "#aa3939" },
```
ascobie: CSS

app/services/monitor-http/core-count-metrics.ts
timotheeguerin: can't really do that here as its for the graph. I am however working on loading themes from a json file so you can use the color variables from javascript and from css

ascobie: Thanks, will remove other related comments

Issue #953, opened 2017-12-05
Create Blender plugin to test BL Custom Protocol


ascobie:
Not really related to Labs per se, but adding here to end-to-end test the custom protocol that Tim is adding support for. 

Simple plugin to test managing and running jobs from an external application.

Existing plugins: 
https://github.com/Azure/azure-batch-apps-blender
https://github.com/annatisch/azure-batch-apps-blender/tree/dev 

Potential URI Scheme

**Just open batchlabs**
ms-batchlabs://

**Open batchlabs at a specific account**
ms-batchlabs://?account=<account-id>

**Open batchlabs at a specific route**
ms-batchlabs://route/<batchlabs-route>

**Open batchlabs at a specific route and account**
ms-batchlabs://route/<batchlabs-route>?account=<account-id>

**Open batchlabs at a specific route**
ms-batchlabs://special-action/params?account=<account-id>
 
**Example:**
ms-batchlabs://route/jobs?account=/subscriptions/21abd678-18c5-4660-9fdd-8c5ba6b6fe1f/resourceGroups/Default-AzureBatch-BrazilSouth/providers/Microsoft.Batch/batchAccounts/prodtest1

PR #1049, opened 2018-02-02
Handle extenal parameters for NCJ

ascobie:
Fix: #953 

- [x] Pass in parameters from an external app
- [x] Update route query parameters on change
- [x] Add validation for file in file group
- [x] Handle the file group prefix so that it doesn't break on reload
- [x] Tests


app/models/ncj-job-template.ts
```diff
@@ -26,8 +26,8 @@ export interface NcjJobTemplate {
 
 export interface NcjPoolTemplate {
     parameters: StringMap<NcjParameter>;
-    pool: NcjJobConfiguration;
-    variables: NcjJobConfiguration;
+    variables?: StringMap<any>;
+    pool: StringMap<string>;
 }
```
ascobie: @timotheeguerin, can you double check this change. it seemed wrong and like a copy/paste issue. didn't matter with binding raw JSON to it, but the tests picked it up.

test/app/components/market/submit/submit-ncj-template.component.spec.ts
```diff
@@ -0,0 +1,215 @@
+import { Component, DebugElement, NO_ERRORS_SCHEMA } from "@angular/core";
```
timotheeguerin: I started moving the test out of the test dir and putting them next to the file they are testing. I feel this makes test writing much easier and force us to have a better structure otherwise folder are overloaded with different files.

For a component I do a folder in this format:

component-name
   component-name.component.ts
   component-name.component.spec.ts
   component-name.html
   component-name.scss
   index.ts

I think we should change the test out of the test dir when we add a new one or update an existing one

Issue #1171, opened 2018-03-14
NCJ Local templates seem to have stopped working


ascobie:
When using this job template: [job.template.json.txt](https://github.com/Azure/BatchLabs/files/1813515/job.template.json.txt)

And i try to change the job id i get this error that crashes the app. 

**Error: There is no FormControl instance attached to form control element with path: 'job -> jobName'**

I did make some changes to this file to support parameters from the blender plugin, but hoping that i didn't affect this.

**Edit:**
As part of changes to allow parameters to be passed it, i added a change event handler to propogate changes between reloads: ```_handleControlChangeEvents```. This seems to not like local templates as it works from any marketplace selected template.

Issue #1129, opened 2018-03-08
Handle multiple folder uploads for a single file group


ascobie:
Handle multiple file and folder uploads when managing file groups.
- [x] Handle both multiple single file and/or multiple folder uploads when creating a file group. 
- [x] Allow the creation of an empty file group from the create FG form.

PR #1175, opened 2018-03-16
File group multi source support

ascobie:
Fix: #1129
Fix: #1171
Fix: #1179

Fix app crash with local templates.
Add support for multiple file group sources when creating or updating.
Fix issue with recent templates and file groups not being prefixed correctly.
Ability to block redirection to pools/jobs page on NCJ submission.
Drag and Drop support for local templates.


data/themes/classic.json
```diff
@@ -36,7 +36,7 @@
   },
   "button": {
     "disabled-text": "#d5d5d5",
-    "disabled-bg": "#f5f5f5"
+    "disabled-bg": "#ededed"
   },
```
ascobie: Made this slightly darker as it was exactly the same color as the NCJ form background and you couldn't see the button when it was disabled. Have seen this in a few places so will sort them out as well.

PR #1222, opened 2018-03-29
Fav account change detection

ascobie:
Favorite account selection was not updating until after you select another account due to observable. 
Change size of 'download directory' string as it was overflowing the background task. (should change css here to auto ellipsis as well rather than wrapping)


app/components/account/browse/account-list/account-list.component.ts
```diff
@@ -75,7 +75,11 @@ export class AccountListComponent extends ListBaseComponent implements OnDestroy
         if (this.isAccountFavorite(accountId)) {
             this.accountService.unFavoriteAccount(accountId);
         } else {
-            this.accountService.favoriteAccount(accountId);
+            this.accountService.favoriteAccount(accountId).subscribe({
+                complete: () => {
```
timotheeguerin: you can just do subscribe(() => {

});

here instead as it should emit multiple values

ascobie: cool, thanks.

ascobie: just changed it and now the favorite reverted back to its old behavior ...

ascobie: changed to this: 

            this.accountService.favoriteAccount(accountId).subscribe(() => {
                this.changeDetector.markForCheck();
            });

and its not being called any more. should i use next: () => {} instead?

PR #1257, opened 2018-04-10
Feature: patch job schedule

### LINKED ISSUE #1170: Support Patching JobSchedules
**Description**
Currently, once a job schedule is created, it cannot be updated. Batch has the ability to patch JobSchedules after they are created.

johnnyzhang82:
Fixed #1170, added edit button for job schedule allows users update job schedule.

image


app/components/job-schedule/action/add/patch-job-schedule-form.component.ts
```diff
@@ -0,0 +1,54 @@
...
+import { SidebarRef } from "@batch-flask/ui/sidebar";
+import { JobScheduleCreateDto, JobSchedulePatchDto } from "app/models/dtos";
+import { JobScheduleService } from "app/services";
+import { JobScheduleCreateBasicDialogComponent } from "./job-schedule-create-basic-dialog.component";
+
+@Component({
+    selector: "bl-patch-job-schedule-form",
+    templateUrl: "job-schedule-create-basic-dialog.html",
+})
+export class PatchJobScheduleComponent extends JobScheduleCreateBasicDialogComponent {
```
timotheeguerin: I added a new functionalities that if you have a title attribute on a form it will use that on the quick access at the top for switching forms

timotheeguerin: you can check, I updated create batch account

PR #1323, opened 2018-05-03
Hotfix: add container configuration in custom image

johnnyzhang82:
Allow users to specify container configuration for custom images.

image


timotheeguerin:
Hhm can you redo those changes to be off stable not master. Here if I change the target to stable it include another PR which should probably not be here

johnnyzhang82:
I reset the commit and remove merge from 'master'

timotheeguerin:
sweet perfect, just merge when ready then

Issue #1364, opened 2018-05-23
Decode URL parameters passed to application


ascobie:
When passing encoded parameters to Labs we need to handle spaces that have been encoded with '+'. This package handles the conversion: 

https://www.npmjs.com/package/decode-uri-component

The built in decodeURIComponent and decodeURI do not handle + encoding for spaces.

PR #1365, opened 2018-05-23
Decode parameters passed into application

ascobie:
Fix: #1364


package.json
```diff
@@ -136,6 +136,7 @@
     "chokidar": "~2.0.0",
     "commander": "~2.13.0",
     "d3": "~4.12.2",
+    "decode-uri-component": "^0.2.0",
```
timotheeguerin: can you change to ~ instead of ^, even though it is locked it shows differently that way when they push a minor version vs a patch

PR #1416, opened 2018-06-08
Upgrade: Webpack 4, Angular 6, Rxjs 6

timotheeguerin:
Those are the main package being updated all 3 have breaking changes.
The largest is `rxjs` which completely changed how to pipe observables. 
There is a package `rxjs-compat` which makes it backward compatible but we should start switching to the new format https://github.com/ReactiveX/rxjs/blob/master/docs_app/content/guide/v6/migration.md

@johnnyzhang82 @ascobie Please check the doc.

tl;dr;
```ts
// This 
obs.map(x=> x*2)
    .filter(x => x > 100)
    .take(1)

// Becomes
import {map, filter, take} from "rxjs/operators";

obs.pipe(
    map(x=> x*2)
    filter(x => x > 100)
    take(1)
)

```

Pros:
* Nicer formating as all operator will be inline
* Allows you to pass custom operators which can make long pipe much cleaner

Cons:
* Simple pipe(One operator) are more tedious
* Always need to import the operator

This PR is blocked right now because angular 6 doesn't support typescript 2.8(They now have a hard check on it which breaks as we already upgraded to typescript 2.8)


app/services/ms-graph/core/aad-graph-http.service.ts
```diff
@@ -35,6 +35,7 @@ export class AADGraphHttpService extends HttpService {
                     .flatMap((accessToken) => {
                         options = this.addAuthorizationHeader(options, accessToken);
                         options = this._addApiVersion(options);
+                        // TODO-TIM check dis (as any)
```
timotheeguerin: Check

src/@batch-flask/extensions/observable.ts
```diff
@@ -1,42 +1,42 @@
-import { AsyncSubject, Observable } from "rxjs";
+// import { AsyncSubject, Observable } from "rxjs";
```
timotheeguerin: Remove

Issue #1379, opened 2018-05-26
Add ability to remove pinned items from drop down


ascobie:
After an entity has been deleted it's no longer possible to remove it from the pinned items collection. Either add a context menu, or just a hover delete button to remove an item.

timotheeguerin:
Let's do middle click to close and add a context menu with delete for accessibility.

Form bookmarks is using the middle click logic

ascobie:
Cool, i also see that the form bookmarks has a delete button. Should we look to add a context menu to that as well so they are the same? or at least if we are going to have a delete button we should make it look the same as the form bookmarks.

timotheeguerin:
Yeah we could add a context menu for bookmarks too

ascobie:
Just so we can make sure they both look the same. Will get Nehal to change the text of the "add-pool" form so it's inline with the other text\.

![image](https://user-images.githubusercontent.com/10934358/42003629-8cbbba82-7ac0-11e8-8913-82885bc94f69.png)

timotheeguerin:
that should be from the title of the form

PR #1473, opened 2018-06-27
Feature: Delete pinned entities from dropdown

nehal-j:
Fix: #1379 

Add ability to remove the pinned favorites from dropdown. 
Changed form titles for pools and certificates


app/components/layout/pinned-entity-dropdown/Untitled-1.txt
```diff
@@ -0,0 +1,20 @@
+//Nehal storage of possibly useful code
+
```
ascobie: Whoops. this was probably my fault. Please remove this file from the Repo :)

app/components/layout/pinned-entity-dropdown/pinned-dropdown.html
```diff
@@ -13,9 +13,12 @@
                 <div class="alias">{{favorite.name || favorite.id}}</div>
                 <div class="type">{{entityType(favorite)}}</div>
             </div>
+            <!-- A green check mark appears if you are on the page corresponding a item in the drop down menu-->
             <div *ngIf="favorite.url === currentUrl" class="extra">
                 <i class="fa fa-check"></i>
             </div>
+            <!-- <bl-button color= "light" icon="fa fa-trash" class="unpin-button" (click)="removeFavorite(favorite)" > unpin </bl-button> -->
```
ascobie: remove commented code

app/components/layout/pinned-entity-dropdown/pinned-dropdown.scss
```diff
@@ -21,4 +21,14 @@ bl-pinned-dropdown {
             font-size: 10px;
         }
     }
+
+    /* Nehal Added the next few LOC..white background when selected
```
ascobie: this will probably be sorted for the final commit so can stay for now.

app/services/pinned-entity.service.ts
```diff
@@ -72,21 +72,22 @@ export class PinnedEntityService {
...
     }
 
-    public getEntityType(entity: NavigableRecord): PinnedEntityType {
-        for (const [type, cls] of pinnedTypeMap) {
-            if (entity instanceof cls) {
-                return type as any;
+    public getEntityType(entity: NavigableRecord | PinnableEntity): PinnedEntityType {
+        for (const [type, cls] of pinnedTypeMap) {
+            if (entity instanceof cls) {
+            return type as any;
```
ascobie: Return needs to be tabbed in 4 spaces. Check the indentation settings.

app/services/pinned-entity.service.ts
```diff
@@ -72,21 +72,22 @@ export class PinnedEntityService {
...
             }
         }
-        return null;
-    }
+        /* casting pinnable as a PinnableEntity */
+        const pinnable = entity as PinnableEntity;
+        return pinnable ? pinnable.pinnableType : null;
+        }
```
ascobie: the } is out of alignment.

app/components/layout/pinned-entity-dropdown/pinned-dropdown.component.ts
```diff
@@ -4,7 +4,9 @@ import { List } from "immutable";
 import { Observable, Subscription } from "rxjs";
 
 import { PinnableEntity, PinnedEntityType } from "@batch-flask/core";
+import { MouseButton } from "@batch-flask/core";
```
ascobie: These 2 imports should be combined into 1.

import { PinnableEntity, PinnedEntityType } from "@batch-flask/core";
import { MouseButton } from "@batch-flask/core";

app/components/layout/pinned-entity-dropdown/pinned-dropdown.component.ts
```diff
@@ -4,7 +4,9 @@ import { List } from "immutable";
...
 import { DropdownComponent } from "@batch-flask/ui";
+import { ContextMenu, ContextMenuItem, ContextMenuService } from "@batch-flask/ui/context-menu";
```
timotheeguerin: this one can also be included from `"@batch-flask/ui`

app/services/pinned-entity.service.ts
```diff
@@ -72,20 +73,22 @@ export class PinnedEntityService {
...
         }
-
         const url = this._fudgeArmUrl(entity);
         const newFavorites = this._favorites.getValue().filter(pinned => pinned.url !== url);
         this._favorites.next(List<PinnableEntity>(newFavorites));
         this._saveAccountFavorites();
     }
 
-    public getEntityType(entity: NavigableRecord): PinnedEntityType {
+    public getEntityType(entity: NavigableRecord | PinnableEntity): PinnedEntityType {
```
timotheeguerin: i'm not sure this will work if entity is a `PinnableEntity` it is just a normal object. `PinnableEntity` is an interface and that method probably won't work for it.

ascobie: yeah i did wonder about that. It does appear to work fine though. The issue we encountered was that the entity from the drop down was a PinnableEntity, not a NavigableRecord. As NavigableRecords are only from the lists.

without this i wasn't able to check pinnable.pinnableType as that doesn't exist on a NavigableRecord. The isFavorite was failing as the type check returned null.

app/components/layout/pinned-entity-dropdown/pinned-dropdown.html
```diff
@@ -3,19 +3,19 @@
...
-            <i class="fa" [ngClass]="entityIcon(favorite)"></i>
-            <div class="main">
-                <div class="alias">{{favorite.name || favorite.id}}</div>
-                <div class="type">{{entityType(favorite)}}</div>
-            </div>
-            <div *ngIf="favorite.url === currentUrl" class="extra">
-                <i class="fa fa-check"></i>
-            </div>
-        </bl-clickable>
+            (contextmenu)="onContextMenu(favorite)">
```
timotheeguerin: order the inputs `[foo]` first then output `(bar)`

app/components/layout/pinned-entity-dropdown/pinned-dropdown.html
```diff
@@ -3,19 +3,19 @@
...
+
+            <span>
+                <i class="entity-type fa" [ngClass]="entityIcon(favorite)" title="{{entityType(favorite)}}"></i>
```
timotheeguerin: [title]="entityType(favorite)"

app/services/pinned-entity.service.ts
timotheeguerin: yeah so it works because of what you did at the bottom. PinnableEntity already has the type as a property. I think you should remove it and use that directly
It doesn't make sense here
```
return pinnable ? pinnable.pinnableType : null;
```

Issue #1449, opened 2018-06-18
BatchLabs: App splited in features that are can be enabled and disabled


timotheeguerin:
Make the app support features that can be enabled and disabled which will not show disabled features.

Examples Features
* `pool`
    * `pool.create`
    * `pool.patch`
* `certificate`
* `gallery`

## Configuration file 
```yaml
{
    "certificate": true,
    "pool": {
        "add": true
    },
    "job": {
        "add": true,
        "view": true
    }
}
```

## Service
```ts
class WorkspaceService { 
    public currentWorkspace: Observable<Workspace>
    selectWorkspace(workspaceId: string);

    public isFeatureEnabled(feature: string): boolean;
    public isFeatureDisabled(feature: string): boolean;
}
```

## Model
```ts
class Workspace {
// Takes the inputs from a file an process

    public isFeatureEnabled(feature: string): boolean;
    public isFeatureDisabled(feature: string): boolean;
}
```

### Directive
Have a directive to use in the templates to make it easier to enable a feature

Name `feature`, `featureEnabled`, `featureOn`?
```html
<div *featureOn="job.view">

</div>


<div *featureOff="job.add">

</div>
```

ascobie:
How about something like this as a first go? completely unsure, but need a way to hide top level navigation as well. I like the separation between viewing and modifying. Though as the first cut, maybe we wont worry about the modification bits at all as these are more permissions based, and expand it into more feature areas.

## Configuration file 
```yaml
{
    "job": {
        "navigate": true,
        "view": {
            "graphs": true,
            "statistics": true,
            "tags": true,
            "asJson": false
        },
        "maintain": {
            "add": true,
            "delete": true,
            "edit": true,
            "addTask": true,
            "terminate": true,
            "clone": true,
            "schedule": true,
            "export": true,
            "pin": true,
        }
    },
    "schedule": {
        "navigate": false,
        "view": {
        },
        "maintain": {
        }
    }
    "pool": {
        "navigate": true,
        "view": {
            "cost": true,
            "tags": true,
            "asJson": true,
            "availableNodes": false,
            "runningTasks": true,
            "appInsights": false,
            "nodes": true,
        },
        "maintain": {
            "add": true,
            "delete": true,
            "addJob": true,
            "resize": true,
            "clone": true,
            "export": true,
            "pin": true,
        }
    },
    "package": {
        "navigate": false,
        "view": {
        },
        "maintain": {
            "add": true,
            "delete": true,
[...]

PR #1477, opened 2018-06-29
Setup workspace and feature switching

ascobie:
Fix: #1449 

- [x] Decide on JSON template structure
- [x] Create master or admin template with all features enabled
- [x] Read workspace templates
- [x] Create workspace selector DDL
- [x] Feature directive for hiding/showing components

Note: ability to override workspace settings can be added in another PR and feature request as this one is already big enough.


ascobie:
So just pass it a workspace template file contents without the user overrides? Then it would parse it and set the Id and name?

Get Outlook for Android<https://aka.ms/ghei36>

app/components/workspace/json-templates/admin-workspace.json
```diff
@@ -0,0 +1,16 @@
+{
+    "metadata": {
```
timotheeguerin: why not move all of those up one level if we have workspace bellow. Also `features` might be better?

app/components/workspace/workspace-selector/workspace-dropdown.component.ts
```diff
@@ -0,0 +1,55 @@
...
+        private workspaceService: WorkspaceService) {
+
+        this.workspaces = this.workspaceService.workspaces;
+        this._subscriptions.push(this.workspaces.subscribe((workspaces) => {
+            console.log("workspaces.subscribe: ", workspaces);
+            this.loaded = true;
+            this.changeDetector.markForCheck();
+        }));
+
+        this._subscriptions.push(workspaceService.currentWorkspace.subscribe((workspace: Workspace) => {
```
timotheeguerin: you probably don't need this one here. This was for accounts

app/services/workspace.service.ts
```diff
@@ -0,0 +1,57 @@
...
+import { List } from "immutable";
+import { BehaviorSubject, Observable } from "rxjs";
+
+// tslint:disable:no-var-requires max-line-length
+const stripJsonComments = require("strip-json-comments");
+
+import { Workspace } from "app/models";
+
+// TODO: can we just read all JSON files in a directory?
+const adminWorkspace = JSON.parse(stripJsonComments(require("app/components/workspace/json-templates/admin-workspace.json")));
```
timotheeguerin: Let's start like that, we'll look into custom user workspace later. You can also remove the stripJsonComments here. This file shouldn't have some

app/components/workspace/json-templates/admin-workspace.json
ascobie: if you are happy with that then we could do. didn't want metadata mixing with the actual control values.

app/components/workspace/workspace-selector/workspace-dropdown.component.ts
ascobie: yeah this was just to test that the current workspace was being set and subscribable. will remove.

.vscode/settings.json
```diff
@@ -26,5 +26,9 @@
   "tslint.alwaysShowRuleFailuresAsWarnings": true,
   "css.validate": false,
   "scss.validate": false,
-  "python.formatting.provider": "yapf"
+  "python.formatting.provider": "yapf",
+  "[json]": {
```
timotheeguerin: can you revert that? We use .editorConfig for those settings and json is set to 2 space on purpose as npm will use 2 space for package.json

ascobie:
Ok, I can revert it. Though viewing the json templates in VS Code is very annoying at 2 spaces. Is there any other way I can change it? I can't read json with 2 spaces.



Get Outlook for Android<https://aka.ms/ghei36>

________________________________

app/app.component.ts
```diff
@@ -51,7 +51,11 @@ export class AppComponent implements OnInit {
     ) {
         this.autoscaleFormulaService.init();
         this.settingsService.init();
+<<<<<<< HEAD
```
timotheeguerin: merge issue

app/components/common/feature-switching/index.ts
```diff
@@ -0,0 +1 @@
+export * from "../../../../src/@batch-flask/ui/workspace/feature-visible.directive";
```
timotheeguerin: remove this file

app/components/job-schedule/details/job-schedule-configuration.component.ts
```diff
@@ -22,8 +24,26 @@ export class JobScheduleConfigurationComponent {
...
 
     private _jobSchedule: JobSchedule;
+    private _sub: Subscription;
+
+    constructor(
+        private changeDetector: ChangeDetectorRef,
+        private workspaceService: WorkspaceService) {
+
+        this._sub = this.workspaceService.currentWorkspace.subscribe((ws) => {
+            if (ws) {
```
timotheeguerin: i feel like we should always assume this is never null. We'll make sure there is always a value in the service

app/components/layout/main-navigation/main-navigation.html
```diff
@@ -2,27 +2,41 @@
...
-        <i class="fa fa-certificate fa-lg"></i><div class="label">Certificate</div>
-    </bl-clickable>
-    <bl-clickable *blHiddenIfNoAccount routerLink="/data" routerLinkActive="active" title="File groups">
-        <i class="fa fa-cloud-upload fa-lg"></i><div class="label">Data</div>
-    </bl-clickable>
-    <bl-clickable *blHiddenIfNoAccount href="#" routerLink="market" routerLinkActive="active" title="Gallery">
-        <i class="fa fa-cubes" aria-hidden="true"></i><div class="label">Gallery</div>
-    </bl-clickable>
+    <div *blHiddenIfNoAccount>
+        <bl-clickable *blFeatureOn="'job.view'" routerLink="/jobs" routerLinkActive="active">
```
timotheeguerin: shouldn't all of those just be `job`?

app/components/workspace/json-templates/end-user-workspace.json
```diff
@@ -0,0 +1,50 @@
+{
+    "id": "end-user.workspace",
+    "displayName": "Basic end-user workspace",
+    "description": "Default end-user workspace that hides a number of admin related features.",
```
timotheeguerin: I would not call admin(Maybe advanced?)

app/components/workspace/json-templates/end-user-workspace.json
```diff
@@ -0,0 +1,50 @@
...
+                    "json": false
+                }
+            },
+            "action": {
+                "clone": false,
+                "exportAsJson": false,
+                "pin": false
+            }
+        },
+        "schedule": {
```
timotheeguerin: feel like we could move all the view one step above. It doesn't really make sense to only disable view

```
{
 "job": {
   "statistics": true,
   "tags": true,
 }
}
```

src/@batch-flask/ui/entity-commands/entity-command.ts
```diff
@@ -17,6 +18,10 @@ export enum EntityCommandNotify {
 }
 
 export interface EntityCommandAttributes<TEntity extends ActionableEntity, TOptions = void> {
+    /**
+     * Name is to be used by feature to show buttons. Or to define keyboard shortcuts
+     */
+    name?: string;
```
timotheeguerin: remove the ? for this one. I added it so it wouldn't break everything but now all commands should have a name(Add to those missing)

src/@batch-flask/ui/workspace/feature-visible.directive.ts
```diff
@@ -0,0 +1,39 @@
...
+    constructor(
+        private templateRef: TemplateRef<any>,
+        private viewContainer: ViewContainerRef,
+        private workspaceService: WorkspaceService,
+    ) {
+    }
+
+    public ngOnInit() {
+        this._subscription = this.workspaceService.currentWorkspace.subscribe((workspace) => {
+            if (this.feature && workspace) {
```
timotheeguerin: you need to listen to the `ngOnChanges` here too otherwise if feature changes then it won't update

src/@batch-flask/ui/workspace/workspace.model.ts
```diff
@@ -0,0 +1,56 @@
...
+            if (definition === true) {
+                return true;
+            } else if (definition === undefined) {
+                return true;
+            } else if (definition === false) {
+                return false;
+            }
+        }
+
+        // TODO: debug - remove
```
timotheeguerin: remove

src/@batch-flask/ui/workspace/workspace.model.ts
```diff
@@ -0,0 +1,56 @@
...
+        if (definition) {
+            // console.log(`traversed feature '${requestedFeature}' to end ... definition tree ->`, definition);
+        }
+
+        // if there are any remaining template definition children after we
+        // have got to the end of the requested feature, we can assume this means
+        // that it should be enabled.
+        return definition ? true : false;
```
timotheeguerin: think this should be definition == false ? false : true

src/@batch-flask/ui/workspace/workspace.service.spec.ts
```diff
@@ -0,0 +1,3 @@
+describe("WorkspaceService", () => {
+    // TODO
```
timotheeguerin: Todo tests

app/components/workspace/json-templates/end-user-workspace.json
ascobie: Action can stay where it is I reckon.

src/@batch-flask/ui/workspace/feature-visible.directive.ts
ascobie: Why would the feature change?

timotheeguerin: why not, might be some computed features

ascobie: Yeah I guess that's true. Future proofing is not a bad idea.

app/components/workspace/json-templates/end-user-workspace.json
nehal-j: After removing view, I need to rename the tags because 'tags' has been used for the summaryTags as well. 
<bl-tags *blFeatureOn="'pool.paper'" summaryTags [tags]="pool.tags" [editable]="true" [save]="updateTags" noTagsMessage="No tags."></bl-tags>

src/@batch-flask/ui/workspace/workspace.service.spec.ts
ascobie: Nehal working on those now.

app/app.component.ts
```diff
@@ -4,7 +4,8 @@ import { DomSanitizer } from "@angular/platform-browser";
 import { Observable } from "rxjs";
 
 import { ActivatedRoute } from "@angular/router";
-import { IpcService } from "@batch-flask/ui";
+import { IpcService, Workspace, WorkspaceService } from "@batch-flask/ui";
+// import { MonacoLoader } from "@batch-flask/ui/editor";
```
timotheeguerin: remove

app/components/data/shared/job-id/job-id.component.ts
```diff
@@ -70,7 +70,12 @@ export class JobIdComponent implements AsyncValidator, ControlValueAccessor, OnD
 
     @autobind()
     private _validateJobUnique(control: FormControl) {
-        return Observable.of(null).debounceTime(250)
+        // on init value is an empty array.
+        if (!control.value || (Array.isArray(control.value) && control.value.length === 0)) {
+            return Observable.of(null);
```
timotheeguerin: change to `of(null)` in rxjs 6

app/components/job/action/job-commands.ts
```diff
@@ -189,4 +197,12 @@ export class JobCommands extends EntityCommands<Job> {
             return Observable.fromPromise(this.fs.saveFile(localPath, content));
```
timotheeguerin: `Observable.fromPromise` change to `from` in genera change all Observable. to the rxjs 6 version

app/app.component.ts
```diff
@@ -5,7 +5,6 @@ import { Observable } from "rxjs";
 
 import { ActivatedRoute } from "@angular/router";
 import { IpcService, Workspace, WorkspaceService } from "@batch-flask/ui";
-// import { MonacoLoader } from "@batch-flask/ui/editor";
 import { PermissionService } from "@batch-flask/ui/permission";
```
ascobie: still don't know how this made its way in. i think you should revert this change.

app/components/data/shared/job-id/job-id.component.spec.ts
```diff
@@ -1,81 +1,86 @@
-import { Component, DebugElement, NO_ERRORS_SCHEMA } from "@angular/core";
```
ascobie: commented out this entire fixture due to it breaking the build with the recent addition of the "timer(delay)". need Tim's assistance to fix it.

ascobie:
Just merged master into this branch. There now seems to be an issue with i18n and it not computing when the component is re-added: 

image

Any ideas how to fix this?

app/services/file/file-navigator/file-tree.model.spec.ts
```diff
@@ -15,6 +15,7 @@ function reprTree(tree: FileTreeStructure) {
 
 function reprNode(node: FileTreeNode, indent = 0) {
     const rows = [];
+    // tslint:disable-next-line:no-unused-variable
     for (const [_, child] of node.children) {
```
ascobie: the _ was failing the build, might have been a package update as this hasn't change for a long time.

timotheeguerin: remove _ just `[,child]`

PR #1511, opened 2018-07-27
Fix for DPs not loading in tab when task change not detected

ascobie:


app/components/task/details/task-dependency-browser/task-dependency-browser.component.ts
```diff
@@ -30,7 +30,8 @@ export class TaskDependencyBrowserComponent implements OnChanges {
             this._tasks.clear();
         }
 
-        if (changes.jobId || ComponentUtils.recordChangedId(changes.task)) {
+        // Task initially loaded with no dedendsOn. Change to task properties was not handled here.
+        if (changes.jobId || ComponentUtils.recordChangedId(changes.task) || changes.task) {
```
ascobie: Tim, is this OK?

timotheeguerin: hhm, I don't think so, here it will recompute every time the task change(which is every 5 seconds as it polls for a new one)

ascobie: ok, need something to trigger the first time its reloaded then. The first time there are no DPs, second time they are there. I could do something with the InitialChange or FirstChange or whatever that property is called. Or add the task change to another conditional that only executes once.

timotheeguerin: yeah so maybe there is a bug in recordChangedId because that should have triggered I believe

app/components/task/details/task-dependency-browser/task-dependency-browser.component.ts
```diff
@@ -30,12 +30,18 @@ export class TaskDependencyBrowserComponent implements OnChanges {
             this._tasks.clear();
         }
 
-        if (changes.jobId || ComponentUtils.recordChangedId(changes.task)) {
+        // Task initially loaded with no dedendsOn. Change to task properties was not handled here.
+        if (changes.jobId || ComponentUtils.recordChangedId(changes.task) || this._hasUnloadedDependencies) {
             this._loaded = 0;
```
ascobie: This only gets called twice now if it needs to be. The first time and the next time if there was a change in the DP property

ascobie:
confirmed that this is now fixed.

Issue #1501, opened 2018-07-18
Add task runtime to task grid


ascobie:
would be very handy to have the task run-time in the main task grid. as soon as the task has started we can start computing it.

PR #1514, opened 2018-07-31
Add run time to the task grid

ascobie:
Fix: #1501

image

Shows success and failure color matching to state column.

### DIFF
--- a/app/components/task/browse/preview/task-runtime.scss
+++ b/app/components/task/browse/preview/task-runtime.scss
@@ -1,0 +1,10 @@
+@import "app/styles/variables";
+
+bl-task-runtime {
+    // inherits .failure and .success from base.scss
+    color: map-get($primary, 400);
+
+    .fa-clock-o {
+        margin-right: 4px;
+    }
+}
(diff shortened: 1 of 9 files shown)


ascobie:
If you want to complete this for release 17.0 before I get a chance to write the tests. I have a TODO to move the remaining test specs into the correct locations and will sort out the tests for task-preview and task-runtime tomorrow.

timotheeguerin:
should you move that just right of the state?

ascobie:
I thought about that a fair bit. But i kind thought that it read better going: created, started, run time, completed. It seemed to have a natural flow to it.

But TBH, i am not too bothered if you think it would be better next to the state.

timotheeguerin:
Well I think as we have color here too that would work better

app/components/task/browse/preview/task-runtime.scss
```diff
@@ -0,0 +1,10 @@
+@import "app/styles/variables";
+
+bl-task-runtime {
+    // inherits .failure and .success from base.scss
+    color: map-get($primary, 400);
```
timotheeguerin: change to `$primary-color`

PR #1527, opened 2018-08-02
Refactor: using getRandomValues for secure-util.ts

johnnyzhang82:
Replacing Math.random() with window.crypto.getRandomValues for generating random index.


src/@batch-flask/utils/secure-utils.ts
```diff
@@ -16,17 +16,23 @@ export class SecureUtils {
...
     public static password(length: number = 15): string {
         const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%^&*()_+~`|}{[]\:;?><,./-=";
         let retVal = "";
-        for (let i = 0, n = charset.length; i < length; ++i) {
-            retVal += charset.charAt(Math.floor(Math.random() * n));
+        // create an array and fill with random integer
+        const randomArray = new Uint32Array(length);
+        window.crypto.getRandomValues(randomArray);
+        for (let i = 0; i < length; i++) {
+            retVal += charset.charAt(randomArray[i] % charset.length);
```
timotheeguerin: can you use something else than appending to a string here for performance(Maybe a buffer or just an array)

PR #1565, opened 2018-08-28
Disallow colon character in storage container names

### LINKED ISSUE #1564: Disallow ':' character in storage container names
File conventions disallow the colon (':') character, but jobs are allowed to come through with these in the name. We should replace all instances with dashes as per convention.

arinehouse:
Closes #1564

### DIFF
--- a/app/utils/storage-utils.spec.ts
+++ b/app/utils/storage-utils.spec.ts
@@ -1,6 +1,6 @@
 import { StorageUtils } from "app/utils";
 
-fdescribe("StorageUtils", () => {
+describe("StorageUtils", () => {
     it("Throws error if no jobId supplied", (done) => {
         StorageUtils.getSafeContainerName(null).catch((error) => {
             expect(error.message).toEqual("No jobId supplied to getSafeContainerName(jobId: string)");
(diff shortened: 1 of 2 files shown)


app/utils/storage-utils.spec.ts
```diff
@@ -5,7 +5,7 @@ describe("StorageUtils", () => {
     let startAndEndRegex: RegExp;
 
     beforeEach(() => {
-        multipleCharRegex = (StorageUtils as any)._regexUnderscoresAndMultipleDashes;
+        multipleCharRegex = (StorageUtils as any)._regexInvalidCharacters;
```
timotheeguerin: is this being used?

app/utils/storage-utils.spec.ts
```diff
@@ -47,9 +49,8 @@ describe("StorageUtils", () => {
     });
 });
 
-function assertName(done: any, jobId: string, expected: string) {
+function assertName(jobId: string, expected: string) {
     StorageUtils.getSafeContainerName(jobId).then((containerName: string) => {
```
timotheeguerin: why did you remove done? this is async no? should probably change to an async function all those tests instead of done anyway

arinehouse: It's just calling `assertName` at the bottom of the file, which isn't async. Each `assertName` call was calling `done()` at the end, and the entire test would pass as soon as the first `assertName` completed, even if any others were failing.

app/utils/storage-utils.spec.ts
arinehouse: It's used in the 4th `it` block. I'm not sure exactly why though, maybe just to validate that the regex is what we expect, without testing the StorageUtils class? I can remove if it's unnecessary, I think it was a direct copy from the portal tests.

PR #1569, opened 2018-08-28
Add upload progress to footer as a percentage

### LINKED ISSUE #1567: Investigate why upload progress is no longer being reported when uploading filegroup data
When uploading files to a file group via the "create file group" form, it used to show the process, i.e. "1 of 5 files uploaded" with a percentage indicator. Now it doesn't .... 

Check that large file upload progress is also shown ... 

**Edit: large file upload progress is also no longer displayed, sigh.**

arinehouse:
Closes #1567


ascobie:
This all looks ok, but will grab the branch and test it. Are there any Unit Tests you can write to validate that it works as expected and truncates long strings, etc ... ?

PR #1742, opened 2018-11-06
move old tests from /test/app/ => /app/

ascobie:
move the old tests from /test/app/ => /app/..
No file changes, just moves.


timotheeguerin:
nice

PR #1787, opened 2018-11-16
Refactor vmsize exclude pattern

### LINKED ISSUE #1783: Refactor VM size picker with filter
- [ ] Using include pattern to filter batch service unsupported VM sizes ([supported vms](https://msazure.visualstudio.com/DefaultCollection/One/_git/Batch-Service?path=%2Fsrc%2Fxstore%2Fwatask%2FCommon%2FManaged%2FSrc%2FWATaskDefs.cs&version=GBPMBackgroundDelete))
- [ ] Added vm size filters (name search, category)
- [ ] Create an advanced filter UI to filter min/max cpu, min/max ram, min/max price range (TBD)

johnnyzhang82:
Fixed #1783 
Change to include pattern and add two basic filters (category and search name)


johnnyzhang82:
image

src/app/models/vm-size.ts
```diff
@@ -50,3 +55,141 @@ export class VmSize extends Record<VmSizeAttributes> {
         super({ ...data, id: data.name && data.name.toLowerCase() } as any);
     }
 }
+
```
timotheeguerin: why is this here? shouldn't that be loaded?

src/app/components/pool/action/add/vm-size-picker/mock-vm-size.ts
```diff
@@ -0,0 +1,419 @@
+// All possible Vms
```
timotheeguerin: what is this file for?

src/@batch-flask/utils/string-utils/string-utils.ts
```diff
@@ -11,6 +11,16 @@ export class StringUtils {
         return new RegExp(`^${wildcard.replace(/\*/g, ".*")}$`, flags).test(str);
     }
 
+    /**
+     * Check whether a given string match a given regex
+     * @param str string to test
+     * @param regex regular expression
+     * @returns Boolean if the string match the regex
+     */
+    public static regexExpTest(str: string, regex: string): boolean {
```
timotheeguerin: seems like a bit useless to have this method no?

src/app/components/pool/action/add/vm-size-picker/vm-size-picker-filter.component.ts
```diff
@@ -0,0 +1,59 @@
...
+    selector: "bl-vm-size-picker-filter",
+    templateUrl: "vm-size-picker-filter.html",
+    providers: [
+        { provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => VmSizePickerFilterComponent), multi: true },
+        { provide: NG_VALIDATORS, useExisting: forwardRef(() => VmSizePickerFilterComponent), multi: true },
+    ],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class VmSizePickerFilterComponent {
+    @Input() public set categoriesDisplayName(names: any) {
```
timotheeguerin: use this instead https://stackoverflow.com/questions/45819123/how-to-loop-over-object-properties-with-ngfor-in-angular

src/app/models/vm-size.ts
johnnyzhang82: This is used for testing or also used when loading failed

src/app/components/pool/action/add/vm-size-picker/vm-size-picker-filter.component.ts
```diff
@@ -0,0 +1,59 @@
...
+    private _propagateChange: (value: VmSizeFilterValue) => void = null;
+
+    constructor(formBuilder: FormBuilder) {
+        this.form = formBuilder.group({
+            category: ["all"],
+            searchName: [],
+        });
+
+        this.form.valueChanges
+        .pipe(debounceTime(400), distinctUntilChanged())
```
timotheeguerin: probably don't need the debounce here as we are not calling server

src/app/models/vm-size.ts
timotheeguerin: not a big fan of this. Loading shouldn't fail. I would better show a text box if vm sizes are not loaded

src/app/models/vm-size.ts
```diff
@@ -1,5 +1,10 @@
 import { Model, Prop, Record } from "@batch-flask/core";
 
+export interface VmSizeFilterValue {
```
timotheeguerin: move this to the filter component file

src/app/services/vm-size.service.ts
```diff
@@ -49,34 +49,35 @@ export class VmSizeService {
     };
 
     private _sizes = new BehaviorSubject<List<VmSize>>(null);
-    private _excludedSizes = new BehaviorSubject<ExcludedSizes>(null);
+    private _includedSizes = new BehaviorSubject<IncludedSizes>(null);
```
timotheeguerin: change type to `BehaviorSubject<IncludedSizes|null>`

do that for the others.  See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html

I enabled for the node environment but want to do everywhere so better start following the right format

src/app/components/pool/action/add/vm-size-picker/vm-size-picker.component.spec.ts
```diff
@@ -32,30 +33,27 @@ describe("VmSizePickerComponent", () => {
...
-                new VmSize({ name: "Standard_A1" } as any),
-                new VmSize({ name: "Standard_A2" } as any),
-                new VmSize({ name: "Standard_A3" } as any),
-                new VmSize({ name: "Standard_C1" } as any),
-                new VmSize({ name: "Standard_C2" } as any),
-            ])),
-        };
+        passVms = [];
+        iaasVms = [];
+        const vmSizeStrings = supportedVms.split("\n").filter(value => !!value.trim());
```
timotheeguerin: that seems a bit unecessary to have this. Why not just have mock data as it was before? It makes it harder to see why test are failing if it loads all that big file

src/app/models/vm-size.ts
johnnyzhang82: We can move this to mock test instead

src/app/components/pool/action/add/vm-size-picker/vm-size-picker-filter.component.ts
```diff
@@ -0,0 +1,59 @@
+import { ChangeDetectionStrategy, Component, EventEmitter, Input, Output, forwardRef } from "@angular/core";
```
timotheeguerin: add test for this file too

src/app/components/pool/action/add/vm-size-picker/vm-size-picker.component.ts
```diff
@@ -191,7 +179,7 @@ export class VmSizePickerComponent implements ControlValueAccessor, OnInit, OnCh
         this.loadingStatus = LoadingStatus.Ready;
         let vmSizes = this._vmSizes.toArray();
         if (this._currentFilter && this._categoryRegex) {
-            if (this._currentFilter.category) {
+            if (this._currentFilter.category && this._categoryRegex[this._currentFilter.category]) {
```
timotheeguerin: change to ` && this._currentFilter.category in this._categoryRegex`

src/app/components/pool/action/add/vm-size-picker/vm-size-picker-filter.component.ts
johnnyzhang82: added

PR #1789, opened 2018-11-19
Auto add default file group output container

ascobie:
User can define a default output container to use when submitting gallery template jobs. Handles one being passed in from a plugin using the ms-batchlabs:// route parameters also.

Fix: #1790

### DIFF
--- a/src/app/components/data/shared/file-group-picker/file-group-picker.component.ts
+++ b/src/app/components/data/shared/file-group-picker/file-group-picker.component.ts
@@ -77,12 +77,26 @@ export class FileGroupPickerComponent implements ControlValueAccessor, OnInit, O
             this.fileGroupsData.params = {
                 storageAccountId,
             };
+
             this.fileGroupsData.setOptions({
                 filter: FilterBuilder.prop("name").startswith(Constants.ncjFileGroupPrefix),
             });
 
-            this.fileGroupsData.fetchAll().subscribe(() => {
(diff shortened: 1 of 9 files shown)


src/app/components/data/shared/file-group-picker/file-group-picker.component.ts
```diff
@@ -77,12 +77,26 @@ export class FileGroupPickerComponent implements ControlValueAccessor, OnInit, O
             this.fileGroupsData.params = {
                 storageAccountId,
             };
+
             this.fileGroupsData.setOptions({
                 filter: FilterBuilder.prop("name").startswith(Constants.ncjFileGroupPrefix),
             });
 
-            this.fileGroupsData.fetchAll().subscribe(() => {
+            this.fileGroupsData.fetchAll().subscribe((groups: List<BlobContainer>) => {
```
timotheeguerin: why do you actually need to add this? Doesn't the sdk will create the file group if it doesn't exists?

ascobie: yea i guess i create it as i didn't trust the CLI would actually always create it, and also for validation. Just seemed easier for me to create it and have validation pass.

timotheeguerin: well I think then it goes against the UI which says that the file group will be created if you give a name that is not found

PR #2011, opened 2019-04-01
Dynamically adjust recent template text based on window size

### LINKED ISSUE #1969: Recent templates from "my library" are a bit too long
Latest insider: 

recent template list for "my library" templates are a little too long. maybe truncate the start of them.

image

zfengms:
fix #1969 Dynamically adjust recent template text based on window size

### DIFF
--- a/src/app/components/gallery/home/recent-template-list/recent-template-list.component.ts
+++ b/src/app/components/gallery/home/recent-template-list/recent-template-list.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnDestroy } from "@angular/core";
+import { Component, HostListener, OnDestroy } from "@angular/core";
 import { ElectronRemote } from "@batch-flask/electron";
 import { DialogService } from "@batch-flask/ui";
 import { NcjTemplateMode } from "app/models";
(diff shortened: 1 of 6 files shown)


src/app/components/gallery/home/recent-template-list/recent-template-list.component.ts
```diff
@@ -23,9 +23,25 @@ export class RecentTemplateListComponent implements OnDestroy {
         private dialogService: DialogService) {
         this._subs.push(this.templateService.recentSubmission.subscribe((value) => {
             this.recentSubmissions = value;
+            this.onResizeEvent(null);
         }));
     }
 
+    @HostListener("window:resize", ["$event"])
```
timotheeguerin: you should be able to do all of that just with css. This looks a bit too complex for what it should be

ascobie: The issue with CSS is that it trims the end of the string, we want the start of the string to be trimmed so we can still see the template name, if we can do this with CSS then go for it. Just not sure it's doable.

timotheeguerin: First yes it is possible to ellipsis at the beginning, we talked with zhiqiang the issue here is the title is not just the filename it contains the entire title shown on the submit form `Run template foo.ts`. Adding ... in the middle depending on what the text start with is quite hacky and will break at some point(Localization for example) For now we'll just make sure the button doesn't show up alone on the new line. Then we can think of a way to change so that only the path(Or action name) is passed to the recent template so that can be stripped

PR #2022, opened 2019-04-16
Update constants to include CS6

bgklein:
Now that this is fully deployed add in constants to mirror other CS versions

Are there tests I need to run or do you trust me?


zfengms:
#2027

PR #2037, opened 2019-05-13
Update auto-user defaults to target pool scope

bgklein:
Switch defaults around, to align with best practices (even though default through REST is different)

### DIFF
--- a/src/app/components/task/base/user-identity/user-identity-picker.component.ts
+++ b/src/app/components/task/base/user-identity/user-identity-picker.component.ts
@@ -14,16 +14,6 @@ interface UserOption {
 
 const defaultUsers = [
     {
-        label: "Task user",
-        scope: AutoUserScope.task,
-        elevationLevel: UserAccountElevationLevel.nonadmin,
-    },
-    {
-        label: "Task user (Admin)",
-        scope: AutoUserScope.task,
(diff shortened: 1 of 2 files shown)


src/app/components/task/base/user-identity/user-identity-picker.component.ts
```diff
@@ -14,23 +14,23 @@ interface UserOption {
 
 const defaultUsers = [
     {
-        label: "Task user",
-        scope: AutoUserScope.task,
+        label: "Pool user",
```
timotheeguerin: are you doing it this way to change the order?

bgklein: Yea figured default should be first? Plus then its alphabetical too which is always a plus :)

timotheeguerin: Yeah make sense, so was the default already pool user? I don't see that changed.

bgklein: The default was 0 in this list, so just changing order and default.

timotheeguerin: sounds good, just fix the test and all good

PR #2054, opened 2019-07-03
Add SharedImageGallery to custom images blade

### LINKED ISSUE #2051: Shared Image Gallery
Please provide an option to use Shared Image gallery from batch explorer with azure pools

### LINKED ISSUE #2053: Subscription List cannot be pulled
**Describe the bug**
Installed the latest version of Batch Explorer. Logged in with my @microsoft.com account. No subs appear in dropdown. Console is showing the following:

image

**To Reproduce**
1. Install Batch Explorer
2. Log into user account

**Expected behavior**
Subscriptions listed.

**Screenshots**
image

**Logs(Click profile > View Logs > app.log, client.log)**
If applicable please attach those log files
image

 - OS: [Windows/Osx/Linux]
 - Version [e.g. 0.15.2]

bgklein:
Fixes #2051

### DIFF
--- a/src/app/components/pool/action/add/os-picker/custom-image-picker/custom-image-picker.component.ts
+++ b/src/app/components/pool/action/add/os-picker/custom-image-picker/custom-image-picker.component.ts
@@ -10,11 +10,10 @@ import { ServerError } from "@batch-flask/core";
 import { LoadingStatus } from "@batch-flask/ui";
 import { ArmBatchAccount, NodeAgentSku, Resource } from "app/models";
 import { BatchAccountService, ComputeService, PoolOsService } from "app/services";
-import { Observable, Subject, concat, forkJoin, of } from "rxjs";
-import { distinctUntilChanged, map, switchMap, takeUntil } from "rxjs/operators";
+import { Subject, of } from "rxjs";
+import { distinctUntilChanged, switchMap, takeUntil } from "rxjs/operators";
 
 import "./custom-image-picker.scss";
-import { image } from "d3";
 
 export interface CustomImageSelection {
     imageId: string;
@@ -93,11 +92,7 @@ export class CustomImagePickerComponent implements OnInit, OnDestroy, ControlVal
                 if (!subscriptionId || !location) {
                     return of([]);
                 }
-                const managedImages = this.computeService.listCustomImages(subscriptionId, location);
-                const sharedImageGalleries = this.computeService.listSIG(subscriptionId, location);
-                return forkJoin(managedImages, sharedImageGalleries, (images, galleries) => {
-                    return [...images, ...galleries];
-                });
+                return this.computeService.listCustomImages(subscriptionId, location);
             }),
         ).subscribe({
             next: (resources) => {
--- a/src/app/components/pool/action/add/os-picker/custom-image-picker/custom-image-picker.component.ts
+++ b/src/app/components/pool/action/add/os-picker/custom-image-picker/custom-image-picker.component.ts
@@ -10,10 +10,11 @@ import { ServerError } from "@batch-flask/core";
 import { LoadingStatus } from "@batch-flask/ui";
 import { ArmBatchAccount, NodeAgentSku, Resource } from "app/models";
 import { BatchAccountService, ComputeService, PoolOsService } from "app/services";
-import { Subject, of } from "rxjs";
-import { distinctUntilChanged, switchMap, takeUntil } from "rxjs/operators";
+import { Observable, Subject, concat, forkJoin, of } from "rxjs";
+import { distinctUntilChanged, map, switchMap, takeUntil } from "rxjs/operators";
 
 import "./custom-image-picker.scss";
+import { image } from "d3";
 
 export interface CustomImageSelection {
     imageId: string;
@@ -92,8 +93,11 @@ export class CustomImagePickerComponent implements OnInit, OnDestroy, ControlVal
                 if (!subscriptionId || !location) {
                     return of([]);
                 }
-
-                return this.computeService.listCustomImages(subscriptionId, location);
+                const managedImages = this.computeService.listCustomImages(subscriptionId, location);
+                const sharedImageGalleries = this.computeService.listSIG(subscriptionId, location);
+                return forkJoin(managedImages, sharedImageGalleries, (images, galleries) => {
+                    return [...images, ...galleries];
+                });
             }),
         ).subscribe({
             next: (resources) => {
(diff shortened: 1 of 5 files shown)


bgklein:
Do not merge as I want to do some more testing and stuffs

bgklein:
image

src/app/components/pool/action/add/os-picker/custom-image-picker/custom-image-picker.component.ts
```diff
@@ -10,10 +10,11 @@ import { ServerError } from "@batch-flask/core";
 import { LoadingStatus } from "@batch-flask/ui";
 import { ArmBatchAccount, NodeAgentSku, Resource } from "app/models";
 import { BatchAccountService, ComputeService, PoolOsService } from "app/services";
-import { Subject, of } from "rxjs";
-import { distinctUntilChanged, switchMap, takeUntil } from "rxjs/operators";
+import { Observable, Subject, concat, forkJoin, of } from "rxjs";
+import { distinctUntilChanged, map, switchMap, takeUntil } from "rxjs/operators";
 
 import "./custom-image-picker.scss";
+import { image } from "d3";
```
bgklein: remove

src/app/services/compute.service.ts
```diff
@@ -89,6 +151,46 @@ export class ComputeService {
         );
     }
 
+    // private _listSIGGalleries(subscriptionId: string, location: string): Observable<Resource[]> {
```
bgklein: remove excess code or keep for utility if we want a seperate blade

src/app/services/compute.service.ts
```diff
@@ -71,7 +75,65 @@ export class ComputeService {
...
+                    reduce((images, response: ArmListResponse<Resource>) => {
+                        return [...images, ...response.value];
+                    }, []),
+                );
+            }),
+            share(),
+        );
+    }
+
+    // private _createSIGResource(gallery: Resource, image: Resource, version: Resource) {
```
timotheeguerin: remove commented code, same above

src/app/models/image-reference.ts
```diff
@@ -11,7 +11,12 @@ export interface CustomImageAttributes {
     virtualMachineImageId: string;
 }
 
-export type ImageReferenceAttributes = MarketImageReferenceAttributes | CustomImageAttributes;
+export interface SharedImageGalleryImageAttributes {
```
timotheeguerin: Isn't it the same as CustomImageAttributes then?

src/app/services/compute.service.ts
```diff
@@ -71,7 +75,65 @@ export class ComputeService {
     public listCustomImages(subscriptionId: string, location: string): Observable<Resource[]> {
         const params = new HttpParams()
             .set("$filter", `resourceType eq '${computeImageProvider}' and location eq '${location}'`);
```
timotheeguerin: if both are used exactly the same you could always do a or query instead of 2 calls?

bgklein: I like that more

bgklein: I'll change

timotheeguerin: the classics vs arm storage accounts listing use the same logic I believe

timotheeguerin: if you want an example

src/app/models/image-reference.ts
bgklein: yea Ill just have them share since its the same. both just use a vitrualmachineimageid.

bgklein:
Should be good once CI passes. Verified it worked. Ultimately SIG is the same as Custom Image just a different image location so I think this should be enough with no additional tests.

Let me know if you want some tho :)

bgklein:
Talked ot Zui will likely to need to update this as SIG can be replicated to regions where the resource doesn't exist in so will need to use their actual endpoints.

src/app/services/compute.service.ts
```diff
@@ -93,22 +127,53 @@ public number;
...
-
-        return this.subscriptionService.get(subscriptionId).pipe(
+        const x = this.subscriptionService.get(subscriptionId).pipe(
             flatMap((subscription) => {
-                return this.azure.get<ArmListResponse>(subscription, resourceUrl(subscriptionId), options).pipe(
-                    expand(obs => {
-                        return obs.nextLink ? this.azure.get(subscription, obs.nextLink, options) : empty();
+                const locationsUri = `subscriptions/${subscription.subscriptionId}/locations`;
+                return this.azure.get<ArmListResponse<ArmLocationAttributes>>(subscription, locationsUri).pipe(
+                    flatMap(response => response.value.map(x => new ArmLocation(x))),
```
timotheeguerin: I don't think this should be a `flatMap` but just a `map` does this work?

bgklein: yea that works

bgklein: forces the next line to be a double map as well though.

src/app/services/compute.service.ts
```diff
@@ -133,8 +133,8 @@ public number;
             flatMap((subscription) => {
                 const locationsUri = `subscriptions/${subscription.subscriptionId}/locations`;
                 return this.azure.get<ArmListResponse<ArmLocationAttributes>>(subscription, locationsUri).pipe(
-                    flatMap(response => response.value.map(x => new ArmLocation(x))),
-                    flatMap(armLocation => {
+                    map(response => response.value.map(x => new ArmLocation(x))),
+                    map(armLocations => armLocations.map(armLocation => {
```
timotheeguerin: this one is good to be `flatMap` or `switchMap`. I think its better to use those than map then mergeAll or switchAll or concatAll

Basically: 
- map: Just take each value emitted by an observable and return a modified value
- flatMap: Take each value emitted by an observable and return another observable(With 1 or multiple values)
- switchMap: Same as flatMap but will cancel if you unsubscribe.

bgklein: Understood, makes sense. Interestingly if we have a `map`  instead of `flatmap` for 
```
map(response => response.value.map(x => new ArmLocation(x))),
```
it requires concatAll to get from an Observable<Observable<Resource[]>> to just a single Observable (due to the inner map for `ArmLocations.map()`).

Can get rid of the mergeAll though.

timotheeguerin: oh is value an observable?

timotheeguerin: ok, I looked at it a bit more. so the first map should stay a `map` the 2nd can be a `switchMap` or `flatMap`

timotheeguerin: also I would move the3 content of that 2nd map into a separate function as its pretty big on its own

PR #2058, opened 2019-07-19
Fix resize error text getting cutoff in banner

bgklein:
image

========>>

image

Don't love the duplication but its either this or hover text, but hover text may have some accessibility concerns? I guess the message could also just be moved to entirely be in the description so only code + fix in banner


bgklein:
Will likely need to expand this to other banners that were thought to have the CSS applied

timotheeguerin:
I would put a title instead so if they hover they can see the full thing as a tooltip

timotheeguerin:
that's a common experience when havin ellipsis

PR #2086, opened 2019-09-17
Change retention time default and eliminate auto-infinity

### LINKED ISSUE #2083: Batch Explorers Unlimited retention time actually sets value to 7d
Due to a change in defaults (previously unlimited) behaviors meant to trigger an unlimited retention time actually set a value of 7d.

bgklein:
Fixes #2083

### DIFF
--- a/src/@batch-flask/ui/duration-picker/duration-picker.component.ts
+++ b/src/@batch-flask/ui/duration-picker/duration-picker.component.ts
@@ -19,7 +19,7 @@ import {
     NG_VALUE_ACCESSOR,
     NgControl,
 } from "@angular/forms";
-import { FlagInput, UNLIMITED_DURATION_THRESHOLD, coerceBooleanProperty } from "@batch-flask/core";
+import { FlagInput, coerceBooleanProperty } from "@batch-flask/core";
 import { FormFieldControl } from "@batch-flask/ui/form/form-field";
 import { SelectComponent } from "@batch-flask/ui/select";
 import { Duration } from "luxon";
@@ -236,7 +236,7 @@ export class DurationPickerComponent implements FormFieldControl<any>,
                     this.invalidTimeNumber = true;
                     return null;
                 } else {
-                    const duration = Duration.fromObject({
+                    return Duration.fromObject({
                         [this.unit]: Number(this.time),
                     });
                 }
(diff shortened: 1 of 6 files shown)


src/app/components/job/action/add/job-manager-task-picker.html
```diff
@@ -25,7 +25,7 @@
             <bl-error controlName="maxTaskRetryCount" code="validateRange">Retry count values can range from -1 to 100</bl-error>
         </bl-form-field>
         <bl-form-field>
-            <bl-duration-picker formControlName="retentionTime" label="Retention time"></bl-duration-picker>
+            <bl-duration-picker formControlName="retentionTime" label="Retention time" [required]="true"[allowUnlimited]="false" [defaultDuration]="7"></bl-duration-picker>
```
timotheeguerin: missing a space before `[allowUnlimited` same for the other files

src/app/components/job/action/add/job-preparation-task-picker.html
```diff
@@ -20,7 +20,7 @@
             <bl-error controlName="maxTaskRetryCount" code="validateRange">Retry count values can range from -1 to 100</bl-error>
         </bl-form-field>
         <bl-form-field>
-            <bl-duration-picker formControlName="retentionTime" label="Retention time"></bl-duration-picker>
+            <bl-duration-picker formControlName="retentionTime" label="Retention time" [required]="true"[allowUnlimited]="false" [defaultDuration]="7"></bl-duration-picker>
```
timotheeguerin: is that all the bl-duration-picker uses? would it make sense to remove unlimited option all together
?

bgklein: Its used for maxwallclocktime as well which makes use of unlimited as a value.

src/@batch-flask/ui/duration-picker/duration-picker.component.ts
```diff
@@ -142,7 +143,13 @@ export class DurationPickerComponent implements FormFieldControl<any>,
     public ngOnChanges(changes) {
         if (changes.allowUnlimited) {
             if (!this.allowUnlimited && this.unit === DurationUnit.Unlimited) {
-                this.unit = DurationUnit.Hours;
+                this.unit = DurationUnit.Days;
+            }
+        }
+        if (changes.defaultDuration) {
+            if (this.defaultDuration) {
+                this.time = this.defaultDuration;
```
timotheeguerin: This would potentially override the currently selected time if the default duration changes. Should probably only do it if the time wasn't changed

bgklein:
Need to fix tests...

PR #2087, opened 2019-09-18
Fix Missing Job Execution Information

### LINKED ISSUE #2088: Start time and End time in Job execution information are empty
**Describe the bug**
Start time and End time under Job execution information show as empty, even though the data is being returned under executionInfo from the API response

**To Reproduce**
1. Go to and existing job
2. Click on Configuration Tab
3. Expand Execution information
4. Execution time is populated (based on the start time), but Start time and End time are both blank

**Expected behavior**
Fields should be populated

**Screenshots**

**Logs(Click profile > View Logs > app.log, client.log)**
If applicable please attach those log files

 - OS: [Windows]
 - Version [2.3.1-stable.378]

bgklein:
image

Fixes #2088

### DIFF
--- a/src/app/components/job/details/job-configuration.html
+++ b/src/app/components/job/details/job-configuration.html
@@ -38,8 +38,8 @@
                 <span *ngIf="job?.executionInfo?.endTime">Completed in: {{ executionInfo.runtime }}</span>
             </div>
 
-            <bl-date-property label="Start time" [value]="job?.executionInfo.startTime"></bl-date-property>
-            <bl-date-property label="End time" [value]="job?.executionInfo.endTime"></bl-date-property>
+            <bl-date-property label="Start time" [value]="job.executionInfo.startTime"></bl-date-property>
+            <bl-date-property label="End time" [value]="job.executionInfo.endTime"></bl-date-property>
             <bl-text-property label="Execution time" [value]="executionInfo.runtime"></bl-text-property>
             <bl-text-property label="Termination reason" [value]="executionInfo.terminateReason"></bl-text-property>
         </bl-property-group>
--- a/src/app/components/job/details/job-configuration.html
+++ b/src/app/components/job/details/job-configuration.html
@@ -38,8 +38,8 @@
                 <span *ngIf="job?.executionInfo?.endTime">Completed in: {{ executionInfo.runtime }}</span>
             </div>
 
-            <bl-text-property label="Start time" [value]="executionInfo.startTime"></bl-text-property>
-            <bl-text-property label="End time" [value]="executionInfo.endTime"></bl-text-property>
+            <bl-text-property label="Start time" [value]="job?.executionInfo.startTime"></bl-text-property>
+            <bl-text-property label="End time" [value]="job?.executionInfo.endTime"></bl-text-property>
             <bl-text-property label="Execution time" [value]="executionInfo.runtime"></bl-text-property>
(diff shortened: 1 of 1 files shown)


bgklein:
Fixes #2088

src/app/components/job/details/job-configuration.html
```diff
@@ -38,8 +38,8 @@
                 <span *ngIf="job?.executionInfo?.endTime">Completed in: {{ executionInfo.runtime }}</span>
             </div>
 
-            <bl-text-property label="Start time" [value]="executionInfo.startTime"></bl-text-property>
-            <bl-text-property label="End time" [value]="executionInfo.endTime"></bl-text-property>
+            <bl-text-property label="Start time" [value]="job?.executionInfo.startTime"></bl-text-property>
```
timotheeguerin: why here and not below then?

bgklein: start and end time aren't on the JobExecutionInfoDecorator. Might be better  to just add them there Ill do that instead

bgklein: Already followed this paradigm for task/other execution times, but really we should just have all the information on the decorator

timotheeguerin: oh I see, actually the decorators are coming from the portal and I don't think they are a good idea to use with angular. Better have a minimal component to display the data a certain way. The decorator don't handle data change very well. I'd recommmened you try to slowly get rid of them. I removed quite a few already

src/app/decorators/job-execution-info-decorator.ts
```diff
@@ -8,13 +8,17 @@ export class JobExecutionInfoDecorator extends DecoratorBase<JobExecutionInforma
...
+    public startTime: string;
+    public endTime: string;
 
     constructor(executionInfo: JobExecutionInformation) {
         super(executionInfo);
 
         this.poolId = this.stringField(executionInfo.poolId);
         this.schedulingError = new FailureInfoDecorator(executionInfo.schedulingError || {} as any);
         this.terminateReason = this.stringField(executionInfo.terminateReason);
+        this.startTime = this.stringField(executionInfo.startTime);
```
timotheeguerin: see my reply to the other comment

src/app/components/job/details/job-configuration.html
timotheeguerin: There is actually a `bl-date-property` should use that instead. This won't work with the timezone feature

bgklein: AH i remember you mentioning that! I'll revert and change to date prop

src/app/components/job/details/job-configuration.html
```diff
@@ -38,8 +38,8 @@
                 <span *ngIf="job?.executionInfo?.endTime">Completed in: {{ executionInfo.runtime }}</span>
             </div>
 
-            <bl-text-property label="Start time" [value]="executionInfo.startTime"></bl-text-property>
-            <bl-text-property label="End time" [value]="executionInfo.endTime"></bl-text-property>
+            <bl-date-property label="Start time" [value]="job?.executionInfo.startTime"></bl-date-property>
```
timotheeguerin: I also don't think you need the `?` after job. This component assume a job is being passed or it should most likely do a check at the top

timotheeguerin: there is a lot of other uses in this file which dont have it so it would crash on those anyway

bgklein: All good I'll change

PR #2091, opened 2019-09-27
Fix bug where required caused save to be disabled even though populated

bgklein:


src/app/components/task/action/add/add-task-form.html
```diff
@@ -47,7 +47,7 @@
                 </div>
                 <div class="form-element">
                     <bl-form-field>
-                        <bl-duration-picker formControlName="retentionTime" label="Retention time" [required]="true" [allowUnlimited]="false" [defaultDuration]="7"></bl-duration-picker>
```
timotheeguerin: Seems like you are removing the functionality instead of fixing it here no?

bgklein: True true.

timotheeguerin:
Kinda confused by what you changed now

src/@batch-flask/ui/duration-picker/duration-picker.component.ts
```diff
@@ -76,7 +76,7 @@ export class DurationPickerComponent implements FormFieldControl<any>,
     @Input() public allowUnlimited: boolean = true;
     @Input() public defaultDuration: string = "";
 
-    @Input() @FlagInput() public required = false;
+    @Input() @FlagInput() public required: boolean | string = false;
```
timotheeguerin: why should this be a string?

timotheeguerin: the `@FlagInput` should also convert the value to `true` or `false` when you use it without a value i.e `<foo required>` instead of `<foo [required]="true">`

src/@batch-flask/ui/duration-picker/duration-picker.component.ts
```diff
@@ -183,12 +191,11 @@ export class DurationPickerComponent implements FormFieldControl<any>,
     }
 
     public validate() {
-        if (this.invalidTimeNumber || this.invalidCustomDuration) {
+        if (this.invalidTimeNumber || this.invalidCustomDuration || (this.required && !this.time)) {
             return {
                 duration: "Invalid",
             };
         }
-        return false;
```
timotheeguerin: what does it return now?

bgklein: Null (https://angular.io/api/forms/Validator#validate)

bgklein: False does the same thing but not returning meets the standard better.

timotheeguerin: well acutally not returning means it return `undefined` you should try to return null explicitly(And may add the return type defined in the doc to that function)

bgklein: Ah yea

src/@batch-flask/ui/duration-picker/duration-picker.component.ts
bgklein: Mainly to mirror https://angular.io/api/forms/RequiredValidator#properties. I'll change it back since we make use of flaginput

PR #2093, opened 2019-09-30
Fix starttask waitforsuccess

cRui861:
Sets waitForSuccess checkbox as checked as default. Tests for waitForSuccess.

### DIFF
--- a/src/app/components/pool/start-task/start-task-picker.component.spec.ts
+++ b/src/app/components/pool/start-task/start-task-picker.component.spec.ts
@@ -1,0 +1,100 @@
+import { CommonModule } from "@angular/common";
+import { HttpClientTestingModule } from "@angular/common/http/testing";
+import { Component, DebugElement, Input } from "@angular/core";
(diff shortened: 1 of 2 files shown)


src/app/components/pool/start-task/start-task-picker.component.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    beforeEach(() => {
+        TestBed.configureTestingModule({
+            imports: [
+                CommonModule,
+                FormModule,
+                FormsModule,
+                ReactiveFormsModule,
+                BrowserDynamicTestingModule,
+                HttpClientTestingModule,
+                ElectronTestingModule,
```
timotheeguerin: I don't think you actually need the `BrowserDynamicTestingModule`, `HttpClientTestingModule` and `ElectronTestingModule` now that you have mocked the services and underlaying components

cRui861: Fixed

PR #2099, opened 2019-10-14
Add ability to filter tasks on execution result

### LINKED ISSUE #2084: Unable to filter for failed tasks
**Describe the bug**
When there is a non-zero "failed" task count on the job pie chart, it is difficult to discover these tasks.

**To Reproduce**
1. Go to a job with failed tasks
2. Click on the filter button

**Expected behavior**
Tickbox labelled "Failed" which shows tasks that have contributed towards this failed task count. Current workaround is to tick complete and exclude 0 exit codes, however that does not show tasks which have been failed and will be retried.

 - OS: Windows
 - Version 2.3.1-stable.378

bgklein:
Fixes #2084


bgklein:
image

src/app/styles/base/advanced-filter.scss
```diff
@@ -21,11 +21,11 @@
     }
 }
 
-bl-adv-filter-statepicker {
```
timotheeguerin: I'd recommend moving stuff out of that file. If you need to modify a style there try to put in in the scss file next to the component instead. Its hard to find otherwise which style apply to what

bgklein: Fair point. I actually had a hard time finding this at first so makes sense to move to its own file (especially since I moved the html to its own file as well)

PR #2115, opened 2019-11-01
Merging Stable into Master

cRui861:
Merging stable into master to keep master up to date.

- Updated stable with two hotfixes


bgklein:
Remember do not squash

PR #2128, opened 2019-11-22
Error Message Duration Exceed Maximum

cRui861:
Set error message for when the duration value chosen for retentionTime and/or maxWallClockTime exceeds the DateTime maximum in the backend.

### DIFF
--- a/src/@batch-flask/core/constants.ts
+++ b/src/@batch-flask/core/constants.ts
@@ -43,13 +43,6 @@ export const DataStoreKeys = {
     selectedWorkspaceId: "selected-workspace-id",
 };
 
-// export const MaxDurations = {
-//     maxDays: 10675199,
-//     maxHours: 256204776,
-//     maxMinutes: 15372286560,
-//     maxSeconds: 922337193600,
-// };
-
 // // server returns "P10675199DT2H48M5.4775807S" for unlimited duration
 export const MaxDurationValue = Duration.fromObject({
     days: 10675199,
--- a/src/@batch-flask/core/constants.ts
+++ b/src/@batch-flask/core/constants.ts
@@ -40,3 +40,11 @@ export const DataStoreKeys = {
      */
     selectedWorkspaceId: "selected-workspace-id",
 };
+
+export const MaxDurations = {
+    // server returns "P10675199DT2H48M5.4775807S" for unlimited duration
+    maxDays: 10675199,
+    maxHours: 256204776,
+    maxMinutes: 15372286560,
+    maxSeconds: 922337193600,
+};
--- a/src/@batch-flask/core/constants.ts
+++ b/src/@batch-flask/core/constants.ts
@@ -1,3 +1,5 @@
+import { Duration } from "luxon";
+
 // tslint:disable:variable-name
 
 export enum MouseButton {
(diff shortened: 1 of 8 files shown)


src/@batch-flask/ui/duration-picker/duration-picker.component.ts
```diff
@@ -246,6 +250,18 @@ export class DurationPickerComponent implements FormFieldControl<any>,
                 if (isNaN(time) || time < 0) {
                     this.invalidTimeNumber = true;
                     return null;
+                } else if (time > Constants.MaxDurations.maxDays && this.unit === "days") {
+                    this.invalidDurationValue = true;
+                    return null;
```
bgklein: collapse to one elseif  since all have same body

src/@batch-flask/ui/duration-picker/duration-picker.component.ts
```diff
@@ -246,6 +250,18 @@ export class DurationPickerComponent implements FormFieldControl<any>,
...
+                } else if (time > Constants.MaxDurations.maxHours && this.unit === "hours") {
+                    this.invalidDurationValue = true;
+                    return null;
+                } else if (time > Constants.MaxDurations.maxMinutes && this.unit === "minutes") {
+                    this.invalidDurationValue = true;
+                    return null;
```
bgklein: move null outside switch statement now that its being reused a bunch

src/common/constants/constants.ts
```diff
@@ -241,3 +241,11 @@ export const TelemetryEvents = {
     applicationStart: "Application start",
     disableTelemetry: "Disable telemetry",
 };
+
+export const MaxDurations = {
```
bgklein: Might be best to make each of these its own variable, then do Hours = Days *24 to make it clear thats where its coming from

src/@batch-flask/ui/duration-picker/duration-picker.component.ts
```diff
@@ -246,6 +250,18 @@ export class DurationPickerComponent implements FormFieldControl<any>,
...
+                } else if (time > Constants.MaxDurations.maxSeconds && this.unit === "seconds") {
+                    this.invalidDurationValue = true;
+                    return null;
                 } else {
                     return Duration.fromObject({
                         [this.unit]: Number(this.time),
```
bgklein: Actually you should be able to make the constant a Duration matching the actual max duration, then simply compare to the Duration generated here

src/@batch-flask/ui/duration-picker/duration-picker.component.ts
cRui861: Shouldn't this `return null` remain inside the default case?

bgklein: Basically something like this
```
        switch (this.unit) {
            case DurationUnit.Custom:
                return this._getCustomDuration(this.time);
            default:
                const time = Number(this.time);
                if (isNaN(time) || time < 0) {
                    this.invalidTimeNumber = true;
                } else {
                    return Duration.fromObject({
                        [this.unit]: Number(this.time),
                    });
                }
        }
        return null;
```

src/@batch-flask/core/constants.ts
```diff
@@ -41,10 +43,17 @@ export const DataStoreKeys = {
...
 };
 
-export const MaxDurations = {
-    // server returns "P10675199DT2H48M5.4775807S" for unlimited duration
-    maxDays: 10675199,
-    maxHours: 256204776,
-    maxMinutes: 15372286560,
-    maxSeconds: 922337193600,
-};
+// export const MaxDurations = {
```
bgklein: delete commented code

src/@batch-flask/ui/duration-picker/duration-picker.component.ts
```diff
@@ -240,34 +240,24 @@ export class DurationPickerComponent implements FormFieldControl<any>,
...
-                } else if (time > MaxDurations.maxHours && this.unit === "hours") {
-                    this.invalidDurationValue = true;
-                    return null;
-                } else if (time > MaxDurations.maxMinutes && this.unit === "minutes") {
-                    this.invalidDurationValue = true;
-                    return null;
-                } else if (time > MaxDurations.maxSeconds && this.unit === "seconds") {
-                    this.invalidDurationValue = true;
-                    return null;
                 } else {
```
bgklein: nit: make this an else if

Issue #2127, opened 2019-11-19
When listing tasks with OData, pagination may not be followed to termination


bgklein:
For a large amount of tasks (1million+), it appears that nextLinks are not being followed to termination.

Note there are thousands of active tasks, but only 4 show up.
![image](https://user-images.githubusercontent.com/18502617/69193978-dd4b3d80-0adc-11ea-96fb-78dbffde569d.png)

Request1:
```
https://bearpscy4.westcentralus.batch.azure.com/jobs/job1511/tasks?$filter=(((state%20eq%20%27active%27)))&$select=id,state,creationTime,lastModified,executionInfo&api-version=2019-08-01.10.0
```

Response1:
```
{
  "odata.metadata":"https://bearpscy4.westcentralus.batch.azure.com/$metadata#tasks","value":[
    {
      "id":"group-b-8116","creationTime":"2019-11-18T02:21:55.6808745Z","lastModified":"2019-11-18T02:21:55.6808745Z","state":"active","executionInfo":{
        "retryCount":0,"requeueCount":1,"lastRequeueTime":"2019-11-18T03:46:46.507445Z"
      }
    },{
      "id":"group-b-8117","creationTime":"2019-11-18T02:21:55.6808745Z","lastModified":"2019-11-18T02:21:55.6808745Z","state":"active","executionInfo":{
        "retryCount":0,"requeueCount":1,"lastRequeueTime":"2019-11-18T03:46:46.494453Z"
      }
    },{
      "id":"group-b-8123","creationTime":"2019-11-18T02:21:55.7008749Z","lastModified":"2019-11-18T02:21:55.7008749Z","state":"active","executionInfo":{
        "retryCount":0,"requeueCount":1,"lastRequeueTime":"2019-11-18T03:46:46.499378Z"
      }
    },{
      "id":"group-b-8131","creationTime":"2019-11-18T02:21:55.6999757Z","lastModified":"2019-11-18T02:21:55.6999757Z","state":"active","executionInfo":{
        "retryCount":0,"requeueCount":1,"lastRequeueTime":"2019-11-18T03:46:46.47972Z"
      }
    }
[...]

PR #2137, opened 2019-12-09
Fix bug in list population where a result with values followed by an …

bgklein:
…empty response with next link would stop further list ops

Fixes #2127

PR #2138, opened 2019-12-11
Update list fetch to follow empty responses

bgklein:
Instead of #2137

### DIFF
--- a/src/@batch-flask/core/data/list-getter/list-getter.ts
+++ b/src/@batch-flask/core/data/list-getter/list-getter.ts
@@ -2,7 +2,7 @@ import { Type } from "@angular/core";
 import { Record } from "@batch-flask/core/record";
 import { List, OrderedSet } from "immutable";
 import { Observable, empty, of } from "rxjs";
-import { expand, map, reduce, share, mergeMap, flatMap } from "rxjs/operators";
+import { expand, map, reduce, share, flatMap } from "rxjs/operators";
 import { DataCache } from "../data-cache";
 import { GenericGetter, GenericGetterConfig } from "../generic-getter";
 import { ContinuationToken, ListOptions, ListOptionsAttributes } from "../list-options";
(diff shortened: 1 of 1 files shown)


src/@batch-flask/core/data/list-getter/list-getter.ts
```diff
@@ -70,38 +70,39 @@ export abstract class ListGetter<TEntity extends Record<any>, TParams> extends G
...
+        const items: TEntity[] = data.map(x => this._createItem(x, params));
+        const keys: OrderedSet<string> = OrderedSet(cache.addItems(items, options.select));
         const token = {
             nextLink,
             params,
             options,
         };
 
-        if (items.size !== 0 && isFirstPage) {
+        if (items.length !== 0 && isFirstPage) {
```
timotheeguerin: are you sure this is correct? Did it change from a Immutable List back to a regular array? Maybe we should add type to response here instead of any

src/@batch-flask/core/data/list-getter/list-getter.ts
```diff
@@ -70,38 +70,39 @@ export abstract class ListGetter<TEntity extends Record<any>, TParams> extends G
             return of(cachedResponse);
         }
 
-        return this.list(params, options).pipe(map(x => this._processItems(cache, x, params, options, true)));
+        return this.list(params, options).pipe(flatMap(x => this._processItems(cache, x, params, options, true)));
```
timotheeguerin: prefer switchMap for this as it will support cancelling. Flatmap will keep making calls until its done. So here if you were to load something and it would take 5min but go away to another page it would keep try loading calls after calls.

src/@batch-flask/core/data/list-getter/list-getter.ts
bgklein: Array.map returns an array

bgklein: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

timotheeguerin: yeah my question was more about the data  being an array vs a List

bgklein: Ah yea its an array 
image

bgklein:
Fixes #2127

PR #2139, opened 2019-12-11
Disable Auto-update

cRui861:
Setting a command line flag to disable auto-update upon application launch.

### DIFF
--- a/src/client/cli.ts
+++ b/src/client/cli.ts
@@ -2,17 +2,19 @@ import * as commander from "commander";
 import { app } from "electron";
 
 export interface BatchExplorerArgs {
-    doNotAutoupdate?: boolean;
+    disableAutoupdate?: boolean;
     updated?: boolean;
     userDataDir?: string;
     ignoreCertificateErrors?: boolean;
     args: string[];
 }
 
+// add user config options (in Settings) to also let the status of the autoupdate to persist - whether that be on or off
+// change naming to "set autoupdate on/off" or "autoupdate set to true/false" -> as Xing/Matt
 export function parseArguments(argv: string[]): BatchExplorerArgs {
     return commander
         .version(app.getVersion())
-        .option("--do-not-autoupdate", "Will not autoupdate the application")
+        .option("--disable-autoupdate", "Disables autoupdate for instance and changes settings to turn off autoupdate")
         .option("--updated", "If the application was just updated")
         .option("--user-data-dir <path>", "Change the user data directory. Used for tests")
         .option("--ignore-certificate-errors", "Ignore https certificate errors")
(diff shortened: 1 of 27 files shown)


timotheeguerin:
Why is this needed? There is a settings that disable auto update on quit already.

src/client/cli.ts
```diff
@@ -11,6 +12,7 @@ export interface BatchExplorerArgs {
 export function parseArguments(argv: string[]): BatchExplorerArgs {
     return commander
         .version(app.getVersion())
+        .option("--do-not-autoupdate", "Will not autoupdate the application")
```
bgklein: should likely make this --autoupdate <true|false> so the user can either enable or disable

src/client/startup.ts
```diff
@@ -68,6 +68,11 @@ export async function startBatchExplorer(args: BatchExplorerArgs) {
         app.commandLine.appendSwitch("ignore-certificate-errors", "true");
     }
 
+    if (args.doNotAutoupdate) {
+        log.warn("Application will not autoupdate");
+        autoUpdater.autoInstallOnAppQuit = false;
```
bgklein: Should likelymake this both set this and update config

src/client/cli.ts
xingwu1: agree. Just use autoupdate, and set default to false.

cRui861: If the flag is set, that means that value is already set to true so in this case, that means autoupdate will be turned off. When the user does not set a flag, then the value is false so in this case, the user did not try to turn off autoupdate.

timotheeguerin:
A little confused by the point of having this flag. Do you expect user to start BatchExplorer from the command line with the flag just to not be able to auto update?

src/client/cli.ts
cRui861: Discussed offline

src/client/cli.ts
```diff
@@ -2,15 +2,19 @@ import * as commander from "commander";
...
     ignoreCertificateErrors?: boolean;
     args: string[];
 }
 
+// add user config options (in Settings) to also let the status of the autoupdate to persist - whether that be on or off
+// change naming to "set autoupdate on/off" or "autoupdate set to true/false" -> as Xing/Matt
 export function parseArguments(argv: string[]): BatchExplorerArgs {
     return commander
         .version(app.getVersion())
+        .option("--disable-autoupdate", "Disables autoupdate for instance and changes settings to turn off autoupdate")
```
bgklein: auto-update is two words so add dashes

cRui861:
@timotheeguerin Yeah, the goal is to allow users to use this flag to disable auto-update when opening the app for an instance. I've noted your other suggestions mentioned offline and will revisit them.

cRui861:
Fixes #2152

PR #2150, opened 2020-02-10
Fixed 3 security vulnerabilities

cRui861:
* Need to fix package tree-kill dependency - will be in a different PR


cRui861:
Fixes #2154

PR #2178, opened 2020-05-08
Autodesk Arnold Meter Change 

cRui861:
Updated Autodesk Arnold meter pricing and units, and added Chaos Group V-Ray RT to list of available software. 
* EULA for V-Ray RT is the same as EULA for V-Ray

Current list of software available:
image

Chaos Group V-Ray RT EULA link:
image

### DIFF
--- a/src/app/services/pricing.service.ts
+++ b/src/app/services/pricing.service.ts
@@ -174,6 +174,11 @@ export class PricingService {
         return pricing;
     }
 
+    /**
+     * Sets the software prices using meter IDs
+     * @param meters RateCardMeter[]
+     * @param pricing BatchPricing
+     */
     private _processSoftwaresPricings(meters: RateCardMeter[], pricing: BatchPricing) {
         for (const meter of meters) {
             if (meter.MeterId in softwareMeterId) {
@@ -184,6 +189,10 @@ export class PricingService {
         }
     }
 
+    /**
+     * Sets billing unit of software prices
+     * @param meter RateCardMeter
+     */
     private _setBillingUnit(meter: RateCardMeter) {
         switch (meter.MeterName.toLowerCase()) {
             case "1 vcpu vm license":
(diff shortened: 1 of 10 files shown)


src/app/services/pricing.service.ts
```diff
@@ -55,15 +55,17 @@ const regionMapping = {
...
     "0ec88494-2022-4939-b809-0d914d954692": BatchSoftwareLicense["3dsmax"],
     "1d3bb602-0cde-4618-9fb0-f9d94805c2a6": BatchSoftwareLicense.maya,
     "e2d2d63e-8741-499a-8989-f5f7ec5c3b3f": BatchSoftwareLicense.vray,
+    "450f680c-b109-486a-8fec-2b9e7ab0fbc9": BatchSoftwareLicense.vrayrt,
 };
 
 @Injectable({ providedIn: "root" })
 export class PricingService {
     public pricing: Observable<BatchPricing>;
+    public billingUnit: string;
```
dpwatrous, line 68: This looks like it should be:

`public billingUnit: SoftwareBillingUnit;`

src/app/services/pricing.service.ts
```diff
@@ -172,16 +174,38 @@ export class PricingService {
...
         }
     }
 
+    /**
+     * Sets billing unit of software prices
+     * @param meter RateCardMeter
+     */
+    private _setBillingUnit(meter: RateCardMeter) {
+        switch (meter.MeterName.toLowerCase()) {
+            case "1 vcpu vm license":
```
dpwatrous, line 198: Do we know how reliable these meter names are?

bgklein: We should probably be using the `meter.Unit`  property to infer this? Looks like their entire API is strings though rather than enums so they could change pretty much whatever they want

cRui861: The `meter.Unit` property only printed out "1 hour" when I tested it. I will keep it as `meter.MeterName` for now unless there are other suggestions.

cRui861:
Fixed #2242

PR #2183, opened 2020-05-18
Dependency Updates

cRui861:


src/@batch-flask/electron/electron-app.service.ts
```diff
@@ -16,7 +16,7 @@ export class ElectronApp {
 
     }
 
-    public getPath(name: string): string {
+    public getPath(name: any): string {
```
paterasMSFT, line 19: Why any?

src/@batch-flask/electron/shell.service.ts
```diff
@@ -13,10 +13,9 @@ export class ElectronShell {
     }
     /**
      * Show the given file in a file manager. If possible, select the file.
-     * @returns Whether the item was successfully shown.
      */
-    public showItemInFolder(fullPath: string): boolean {
-        return this._shell.showItemInFolder(fullPath);
+    public showItemInFolder(fullPath: string): void {
+        this._shell.showItemInFolder(fullPath);
```
paterasMSFT, line 18: Does this no longer return false if unsuccessful?

src/@batch-flask/electron/electron-app.service.ts
cRui861: Line 20 `getPath()` was running into an error when `name` was set to type `string` (even though its signature wanted a `string`) so I changed it to `any` for now.

src/@batch-flask/electron/shell.service.ts
cRui861: Yup

src/@batch-flask/electron/electron-app.service.ts
paterasMSFT: Clunky though it is, we should adhere to the string literal union type they're using as an enum:

```
public getPath(name:  'home' | 'appData' | 'userData' | 'cache' | 'temp' | 'exe' | 'module' | 'desktop' | 'documents' | 'downloads' | 'music' | 'pictures' | 'videos' | 'logs' | 'pepperFlashSystemPlugin'): string {
```

Otherwise, if someone passes something else in, we'll get an error at run-time.

PR #2187, opened 2020-06-10
Supports spaces in paths

gingi:


scripts/install/install-windows.ps1
```diff
@@ -122,8 +122,8 @@ function install-python-dependencies() {
         add-failure "Invalid version of python installed. Needs python >= 3.5. You can either have globably available in the path as python, python3 or set the BL_PYTHON_PATH environment variable."
     }
 
-    Write-Host "$python $root/scripts/install/install-python-dep.py"
-    cmd /c "$python $root/scripts/install/install-python-dep.py"
+    Write-Host "$python `"$root/scripts/install/install-python-dep.py`""
```
dpwatrous, line 125: The quoting on this and the next line looks wrong to me.

scripts/install/install-windows.bat
```diff
@@ -1,2 +1,2 @@
 @echo off
-PowerShell.exe -ExecutionPolicy Bypass -Command "%~dp0\install-windows.ps1 %*"
+PowerShell.exe -ExecutionPolicy Bypass -Command "& '%~dp0/install-windows.ps1' %*"
```
dpwatrous, line 2: Won't the & here make the script non-blocking? (Excuse my lack of Powershell expertise)

bgklein: Thats what you would think, but its powershell. Basically just means run the command in powershell, equivalent roughly to  Invoke-Expression

scripts/install/install-windows.ps1
dpwatrous: @gingi This PR looks fine to me minus the strange quoting here - mind taking a look?

gingi:
Abandoned.

PR #2195, opened 2020-07-09
Update to Electron 9 and Spectron 11

dpwatrous:


dpwatrous:
Looks like there's an issue with Karma and Electron 9 which is causing the tests to hang. I was able to reproduce it with minimal code. Opened an issue [here](https://github.com/twolfson/karma-electron/issues/47)

PR #2198, opened 2020-07-14
BE Download Page - Fix image-alt and link-name issues

cRui861:
Added alternative text for links and images for screen readers.
* Small fix: color contrast issues


cRui861:
Fixed #2243

PR #2199, opened 2020-07-14
Fixed Windows/Linux packaging issues

dpwatrous:
Handles 2 breaking changes from the recent electron-builder update:

- The API for creating a standalone executable in electron-builder changed.
- The Linux AppImage binary filename changed.


cRui861:
Fixed #2244

PR #2209, opened 2020-08-20
Dependency Updates

cRui861:
Updated some dependencies and fixed minor build issue.


src/@batch-flask/utils/date-utils/date-utils.ts
```diff
@@ -111,7 +111,7 @@ export class DateUtils {
         if (!(date instanceof DateTime)) {
             date = DateTime.fromJSDate(date);
         }
-        return date.toLocaleString(DateTime.DATETIME_FULL_WITH_SECONDS);
+        return String(date.toLocaleString(DateTime.DATETIME_FULL_WITH_SECONDS));
```
dpwatrous, line 114: This string conversion looks redundant to me

cRui861: Yeah I actually agree. The build errors I saw before making this change were: Type 'string | null' is not assignable to type 'string'. Type 'null' is not assignable to type 'string'. I could change the return types to be `string | null` instead, would that be better?

dpwatrous: Looks good!

dpwatrous: Looks like this is only failing in Node.js because that's where strict null checking is enabled in tsconfig.node.json

PR #2216, opened 2020-09-10
Include the default Electron User-Agent string

### LINKED ISSUE #2165: BatchExplorer does not work with Okta
Our Azure access is protected by Okta and requires the selection of a certificate.  This works fine with web access/browsers, the Microsoft Teams client, and Office365, etc., but fails in BatchExplorer.   You can find more about this here:

https://help.okta.com/en/prod/Content/Topics/Mobile/Okta_Mobile_Device_Trust_Windows-desktop.htm

I think the issue is that the prompt for the certificate that is normally presented is blocked or canceled.  Fixing this would allow tools such as these to work in more corporate environments.   

Here is what the cert selector looks like when logging into Teams:

image

It looks different in Edge Chromium.

I'm going to try setting my group policy to auto choose the cert for both IE and Chromium to see if that helps.

Any other suggestions welcomed.

Thanks,
James

### LINKED ISSUE #2212: Impossible to verify account and complete login
More information is required to verify the account:
image
but in the next screen of "Additional security verification", even after filling in all fields, I can't click "Next" and proceed with any of the options:
image
image

This goes on even after uninstalling and reinstalling the application, also when deleting the `C:\Users\darispol\AppData\Roaming\BatchExplorer` folder.

Thanks for the help,
Daniele

dpwatrous:
Fixes #2212 and Fixes #2165 

The User-Agent string being sent was exposing a bug in AAD auth when setting up 2-factor verification. Including the default Electron User-Agent string fixed this issue.

### DIFF
--- a/src/client/core/batch-explorer-application.ts
+++ b/src/client/core/batch-explorer-application.ts
@@ -27,7 +27,7 @@ import { MainWindowManager } from "./main-window-manager";
 
 const osName = `${os.platform()}-${os.arch()}/${os.release()}`;
 const isDev = ClientConstants.isDev ? "-dev" : "";
-const userAgent = `(${osName}) BatchExplorer/${ClientConstants.version}${isDev}`;
+const userAgentSuffix = `(${osName}) BatchExplorer/${ClientConstants.version}${isDev}`;
 
 export enum BatchExplorerState {
     Loading,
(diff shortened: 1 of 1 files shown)


bgklein (approved):

bgklein:
Also Fixes #2165

dpwatrous:
Good catch - I totally missed that this was basically the fix you had for Okta!

PR #2227, opened 2020-10-16
Replace hyphens in the build number with underscores

dpwatrous:
When manually building a PR branch, underscores would cause the RPM
build to fail


paterasMSFT (approved):

cRui861:
Fixed #2249

PR #2228, opened 2020-10-16
Updated localization CI user

dpwatrous:
This fixes the localization CI pipeline


dpwatrous:
It would be good to figure out if we can use the Github app auth to automate these commits, but in the meantime this gets the localization pipeline back to working condition using my PAT.

cRui861 (approved):

PR #2230, opened 2020-10-22
Variable slot pricing - update and fix breaking API changes

cRui861:
Updates for variable slot pricing feature in BatchService API version [2020-09-01.12.0](https://docs.microsoft.com/en-us/rest/api/batchservice/batch-service-rest-api-versioning#version-2020-03-01110): 
* Replace property `maxTasksPerNode` with `taskSlotsPerNode` on the pool
* Change the response type of `GetTaskCounts` to return `TaskCountsResult`, which is a complex object containing the previous `TaskCounts` object and a new `TaskSlotCounts` object providing similar information in the context of slots being used
* Add property `requiredSlots` to the task allowing user to specify how many slots on a node it should take up


src/app/components/job/details/job-progress-status/job-progress-status.component.spec.ts
```diff
@@ -50,13 +50,22 @@ describe("JobProgressStatusComponent", () => {
...
                     completed: 8,
                     active: jobId === "large-job" ? 234000 : 12,
                     failed: 2,
                     succeeded: 6,
-                }));
+                });
+                const taskSlotCounts = new JobTaskSlotCounts({
+                    running: 3,
+                    completed: 10,
+                    active: jobId === "large-job" ? 234000 : 12,
```
paterasMSFT, line 63: Let's store this result in a local variable to to avoid duplicating it.

src/app/models/job-task-slot-counts.ts
```diff
@@ -0,0 +1,25 @@
...
+    @Prop() public running: number = 0;
+    @Prop() public completed: number = 0;
+    @Prop() public succeeded: number = 0;
+    @Prop() public failed: number = 0;
+
+    public total: number;
+
+    constructor(data: JobTaskSlotCountsAttributes = {}) {
+        super(data);
+        this.total = this.active + this.running + this.completed;
```
paterasMSFT, line 23: Should this be either a read-only, calculated getter, or a getter with a private setter?

paterasMSFT: In other words, should it change and should we allow the consumer to be the one that changes it?

cRui861: I'm not too sure. I modeled this after the `job-task-counts.js` model to help create this one. If I'm understanding this correctly, these need to be public setters due to the deserialization that happens from this mapping call in the `getTaskCounts` function in this service: https://github.com/Azure/BatchExplorer/blob/2219e31cb8e623d3ce61d8a9784de0e99c28090b/src/app/services/azure-batch/job/job.service.ts#L138

bgklein:
Only glimpsed over the code, but does this address the pool view (of nodes in the pool) where currently running tasks are displayed as a percentage of the VM as green?

cRui861:
> Only glimpsed over the code, but does this address the pool view (of nodes in the pool) where currently running tasks are displayed as a percentage of the VM as green?

There are plans to add the task slot counts in the heatmap UI within the Pool blade but it'll probably come in a separate PR.

src/app/components/job/action/add/job-manager-task-picker.html
```diff
@@ -28,6 +28,13 @@
             <bl-duration-picker formControlName="retentionTime" label="Retention time" [allowUnlimited]="false" [defaultDuration]="7" required></bl-duration-picker>
         </bl-form-field>
     </div>
+    <div class="form-element">
+        <bl-form-field>
+            <input blInput formControlName="requiredSlots" placeholder="Required slots">
+        </bl-form-field>
+        <bl-error controlName="requiredSlots" code="required">Required slots is a required field</bl-error>
+        <bl-error controlName="requiredSlots" code="min">Required slots value must be greater than 0</bl-error>
```
zfengms, line 36: Do we enforce only integer is specified for requiredSlots? We might want to make the description more accurate like "greater than or equal to 1".

src/app/components/job/action/add/job-manager-task-picker.component.ts
```diff
@@ -34,6 +36,7 @@ export class JobManagerTaskPickerComponent extends JobTaskBaseComponent {
         this._baseFormControls["displayName"] = [null];
         this._baseFormControls["killJobOnCompletion"] = [true];
         this._baseFormControls["runExclusive"] = [true];
+        this._baseFormControls["requiredSlots"] = [1, [Validators.required, Validators.min(1)]];
```
zfengms, line 39: We may want to add a max value validator based on pool taskSlotsPerNode settings, but it can be done in another PR.

zfengms (approved):

src/app/components/job/action/add/job-manager-task-picker.html
cRui861: I agree this change would be better, I'll update that.

src/app/components/job/action/add/job-manager-task-picker.component.ts
cRui861: The max value validator based on pool taskSlotsPerNode would be based on the VM sizes chosen for that specific pool right? Do you know how that is determined? Like if there is information that I can pull that has a correspondence between the VM size and the max value for the taskSlotsPerNode?

zfengms: The max value for taskSlotsPerNode is 4 x number of VM cores, when vmSize is not specified, the max value is 256

cRui861:
Fixed #2251

PR #2237, opened 2021-01-06
Fix occasional MacOS freezing issue on application startup

dpwatrous:
This seems to have been caused by a bug in chokidar: https://github.com/paulmillr/chokidar/issues/434

Also included some development process improvements:

- Better error handling when the main window fails to load
- A new VSCode launch profile for debugging both the main and renderer processes at once
- Added pyenv's local .python-version file to .gitignore


cRui861 (approved):

cRui861:
Fixed #2254

PR #2271, opened 2021-02-02
Addresses high-severity accessibility issues.

gingi:
Accessibility Milestone 1

### DIFF
--- a/src/@batch-flask/ui/property-list/table-property/table-property.html
+++ b/src/@batch-flask/ui/property-list/table-property/table-property.html
@@ -13,8 +13,19 @@
             </tr>
         </table>
         <div class="expand-container" *ngIf="canExpand || canCollapse">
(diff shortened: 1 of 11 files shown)


src/@batch-flask/ui/property-list/table-property/table-property.html
```diff
@@ -13,8 +13,20 @@
...
-            <bl-button type="round" color="light" class="collapse" *ngIf="canCollapse" (click)="expanded = false" aria-expanded="true"><i class="fa fa-caret-up"></i></bl-button>
+            <bl-button
+                [type]="expanded ? 'round' : 'wide'"
+                color="light"
+                [attr.class]="expanded ? 'expand' : 'collapse'"
+                [attr.aria-expanded]="expanded"
+                [ngSwitch]="expanded"
+                (click)="toggleExpand($event)"
+                (keydown)="toggleExpand($event)"
+            >
```
cRui861, line 24: NIt: would it be better if we just moved this `>` behind `(keydown)` instead of its own line?

cRui861 (approved):

PR #2275, opened 2021-02-04
Fixes task download button

gingi:


dpwatrous (commented):
Good catch!

Not a blocker, but mind changing _saveFile() to take a `string` instead of `any`? If it had been typed we would have caught this on upgrading Electron.

dpwatrous:
Looks like there's also a unit testing issue (the call is probably just being mocked incorrectly now)

gingi:
> Looks like there's also a unit testing issue (the call is probably just being mocked incorrectly now)

Fixed. Had to make several adjustments for promise-based testing.

dpwatrous (approved):

PR #2292, opened 2021-03-06
Added markdown linting

dpwatrous:
Also fixed up all markdown files so there are no linting errors. Uses https://github.com/DavidAnson/markdownlint


gingi (approved):
Love it.

dpwatrous:
Looks like we had some ES6 polyfills for unit tests that relied on a transitive dependency. If the CI tests pass without them then I think we can safely say they're not needed.

dpwatrous:
One more thing to point out on this PR: I found some references to private NPM repos in package-lock.json and just regenerated it. I did it with NPM 7, so this does upgrade package-lock to v2.

That said, we're already switched the package-lock files for the portal so IMO we might as well do it for Batch Explorer.

gingi (approved):

PR #2295, opened 2021-03-10
Simplified ESLint

gingi:
Extracts a common eslintrc with pared down rules, and adds BatchExplorer-specific rules (e.g., angular templates).


src/app/models/workspace-definition.ts
```diff
@@ -89,5 +89,4 @@ export interface DataFeatures extends CommonFeatures {
     };
 }
 
-export interface GalleryFeatures extends CommonFeatures {
-}
+export type GalleryFeatures = CommonFeatures;
```
cRui861, line 92: I didn't realize we had so many empty interfaces. Why is it better for it to be changed this way?

cRui861 (approved):

PR #2298, opened 2021-03-12
Fix InvalidApplicationLicense error when selecting vray-rt software

cRui861:
Users now can select one option for vray and vray-rt:
image


gingi (approved):

src/common/constants/constants.ts
```diff
@@ -103,7 +103,10 @@ export const ApiVersion = {
     armClassicStorage: "2016-11-01",
     armStorage: "2016-12-01",
     armBatch: "2019-08-01",
-    compute: "2019-03-01",
+    compute: {
+        default: "2019-03-01",
+        skus: "2019-04-01",
```
zfengms, line 108: Any reason we can't remove default and use 2019-04-01 for all compute api calls?

zfengms (approved):

src/common/constants/constants.ts
cRui861: Dave and I took a look at this and it seems like sometimes the APIs within the Compute API sometimes has different versioning so we did this to modularize and for more specific testing.

PR #2321, opened 2021-05-10
Migrates ADAL to MSAL

gingi:
Uses the MSAL library (`@azure/msal-node`) to manage access tokens.

Rather than requesting and managing tokens under various flows (caching, token refresh, etc.), a new `AuthProvider` class depends on MSAL for these features.

This is a first implementation, relying on default behavior (such as native caching instead of more persistent caching), but should at the very least mimic the current user experience. The `msal-node` library still has some outstanding issues (such as incomplete multi-tenant support), but this logic should be encapsulated within `AuthProvider` as implementation details.

Includes various other minor changes and improvements.

* Updates ARM API version, allowing tenant details to be fetched in one request (instead of two)
* Updated test utilities
* Updates debugger settings
* Fixes minor tooltip errors


src/app/services/aad/auth.service.spec.ts
```diff
@@ -110,31 +110,35 @@ describe("AdalService spec", () => {
...
-        it("it calls again the main process if previous call returned an error", async () => {
+        it("calls again the main process if previous call returned an error", async () => {
             remoteSpy.send = jasmine.createSpy("send").and.returnValues(
                 Promise.reject("some-error"),
                 Promise.resolve(token1),
             );
             try {
                 await service.accessTokenDataAsync(tenant1, resource1);
-                fail("Shouldn't have succeeded");
+                fail("First call to accessTokenDataAsync houldn't have succeeded");
```
paterasMSFT, line 120: Typo: "houldn't"

src/app/services/subscription/subscription.service.ts
```diff
@@ -9,12 +9,11 @@ import { List, Set } from "immutable";
 import { AsyncSubject, BehaviorSubject, Observable, Subject, combineLatest, empty, forkJoin } from "rxjs";
 import {
     distinctUntilChanged, expand, filter, first, flatMap, map,
-    publishReplay, reduce, refCount, shareReplay, switchMap, takeUntil,
+    publishReplay, reduce, refCount, shareReplay, switchMap, takeUntil, tap,
```
paterasMSFT, line 12: Is tap being used?

src/client/core/aad/auth-provider.ts
```diff
@@ -0,0 +1,113 @@
...
+                await client.acquireTokenByCode({ ...authRequest, code });
+            if (result) {
+                this._accounts[tenantId] = result.account;
+            }
+            return result;
+        }
+    }
+
+    public logout(): void {
+        this._removeAccount();
```
paterasMSFT, line 69: logout removes the client and account from their respective caches, but is it possible something could be making calls using an instance previously fetched after the user thought they were logged out?

src/client/core/aad/authentication/authentication.service.spec.ts
```diff
@@ -32,42 +37,32 @@ describe("AuthenticationService", () => {
         let result: AuthorizeResult | null;
         let error: AuthorizeError | null;
         let promise;
-        beforeEach(() => {
+        beforeEach(async () => {
             result = null;
             error = null;
             const obs = userAuthorization.authorize("tenant-1");
             promise = obs.then((out) => result = out).catch((e) => error = e);
```
paterasMSFT, line 44: Should this be await obs instead of obs.then?

src/client/core/aad/authentication/authentication.service.ts
```diff
@@ -61,28 +61,42 @@ export class LogoutError extends SanitizedError {
...
+    }
+
+    /**
+     * Authorizes a resource
+     *
+     * @param tenantId The tenant ID
+     * @param resourceURI The resource URI
+     * @returns a promise with AuthorizedResult
+     */
+    public async authorizeResource(tenantId: string, resourceURI: string):
```
paterasMSFT, line 93: I don't see an await here, either. I didn't think you could use async by explicitly returning the promise object, but from the below I'm guessing I'm wrong about that.

src/common/constants/constants.ts
```diff
@@ -99,7 +99,7 @@ export const localStorageKey = {
 };
 
 export const ApiVersion = {
-    arm: "2016-09-01",
+    arm: "2020-01-01",
```
paterasMSFT, line 102: Should this be 2021?

src/client/core/aad/auth-provider.ts
gingi: I'm not sure I totally understand what you mean by "instance previously fetched", but `AuthProvider.logout()` simply resets its own conception of a logged-in user, and cannot ensure that the credentials are removed from the entire system (e.g., browser cache). Subsequent access token fetches will require interactive authentication. The authentication service also clears browser cookies when calling `AuthProvider.logout()` ([relevant lines](https://github.com/Azure/BatchExplorer/blob/9a037200750fd27cb0d8931e97fca26c64c29d99/src/client/core/aad/authentication/authentication.service.ts#L119-L120))

src/client/core/aad/authentication/authentication.service.spec.ts
gingi: I had thought that too. But some of the unit tests need to set up certain expectations before the promise is resolved.

src/client/core/aad/authentication/authentication.service.ts
gingi: The `async` is extraneous because of the implementation (but it's not wrong, since the function returns a promise). But the other public `authorize*` methods are async so this makes it consistent and also allows for future modifications.

src/common/constants/constants.ts
gingi: `2020-01-01` is the latest ARM API version.
https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list

paterasMSFT (approved):

PR #2324, opened 2021-05-26
AB#459: Added 'aria-selected' tag for clickable elements

skapur12:
aria-selected

This screenshot displays how the **aria-selected** tag for the 3ds Max button is set to 'true' (because the user has selected that button), whereas for all other buttons on the page, the tag is set to 'false.'


src/@batch-flask/ui/buttons/clickable/clickable.component.ts
```diff
@@ -95,6 +104,7 @@ export class ClickableComponent implements OnChanges, OnDestroy {
 
     @HostListener("click", ["$event"])
     public handleClick(event: Event) {
+        this.selected = true;
```
cRui861, line 107: What happens when multiple elements on the same page get selected?

skapur12: They all display the 'selected' attribute

cRui861: If each of the options were selected at one point and now all of them display 'selected', what if the user wanted to go through and see which options were available before going back and choosing one? How would they know which one is actively being selected on the page?

skapur12: I now see that I misinterpreted what exactly the ‘selected’ feature was supposed to do. I just fixed it so that **only** the selected element has the value _true_ for aria-selected and all other elements have a value of _false_ for aria-selected. I will update this PR shortly.

src/@batch-flask/ui/buttons/clickable/clickable.component.ts
```diff
@@ -31,6 +31,7 @@ export class ClickableComponent implements OnChanges, OnDestroy {
      */
     @Input() public disabled = false;
 
+
```
cRui861, line 34: Nit: remove extra line

cRui861 (dismissed):

cRui861:
Make sure to squash and merge when merging into master

cRui861 (approved):

cRui861:
Included in #2358

PR #2339, opened 2021-06-15
Publish script bumps version post-release

### LINKED ISSUE #2338: Bump version after release
Edit release pipeline such that the insider build always has the version of the next release.

gingi:
* Creates merge branch for bumping version in package.json and merging stable back into master (subsequent manual step).
* Manually bumped version for insider builds.

Closes #2338

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "name": "Microsoft Corporation",
     "email": "batchexplorer@microsoft.com"
   },
-  "version": "2.10.0",
+  "version": "2.11.0",
   "main": "build/client/main.prod.js",
   "scripts": {
     "ts": "ts-node --project tsconfig.node.json --files",
--- a/scripts/publish/publish.ts
+++ b/scripts/publish/publish.ts
@@ -7,6 +7,8 @@ import {
     createIssue, createPullRequest, getMilestone, githubToken, listMilestoneIssues, listPullRequests,
 } from "./github-api";
 
+const MAIN_BRANCH = "master";
+
 const root = path.resolve(path.join(__dirname, "../.."));
 const allMessages: string[] = [];
 const repoName = "Azure/BatchExplorer";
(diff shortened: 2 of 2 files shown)


scripts/publish/publish.ts
```diff
@@ -99,12 +104,20 @@ async function switchToNewBranch(branchName: string) {
...
-    success(`Updated version in package.json to ${version}`);
+    const currentBranch = getCurrentBranch();
+    const nextVersion = calcNextVersion(version);
+    const bumpBranch = `release/bump-${nextVersion}`;
+    await run(`git checkout master`);
+    await run(`git checkout -b ${bumpBranch}`);
+    await run(`npm version --no-git-tag-version --allow-same-version ${nextVersion}`);
+
+    await run(`git commit -am "Bump version to ${nextVersion}"`);
+    await run(`get checkout "${currentBranch}"`);
```
dpwatrous, line 115: Should we be pushing the version bump branch here? What happens if the version bump branch already exists?

Also, looks like this should be `git checkout -b ${currentBranch}`

dpwatrous (changes requested):

scripts/publish/publish.ts
gingi: Um, it doesn't have to have `-b` because the branch already exists. But the command is definitely `git` and not `get`. Whoops.

And yes, we should be pushing to remote. I'm adding a `--force-with-lease` in case the branch already exists.

package.json
```diff
@@ -16,7 +16,7 @@
     "name": "Microsoft Corporation",
     "email": "batchexplorer@microsoft.com"
   },
-  "version": "2.10.0",
+  "version": "2.11.0",
```
cRui861, line 19: Why does this version need to be moved to 2.11.0?

gingi: The version on `master` will always be the next version to be released.

scripts/publish/publish.ts
dpwatrous: Changes look good to me

dpwatrous (approved):

cRui861 (approved):

PR #2345, opened 2021-06-30
Fix tooltip doesn't close with esc key

cRui861:
Tooltip now collapses when esc key is pressed during hover.


gingi (approved):

cRui861:
Included in #2358

PR #2349, opened 2021-07-03
Fix sort icon luminosity contrast ratio too low

cRui861:
The sort icon color contrast (5.7:1) is the same as the label as shown in the screenshot below. When clicked, it turns blue.
image

### DIFF
--- a/src/@batch-flask/ui/table/table.scss
+++ b/src/@batch-flask/ui/table/table.scss
@@ -174,7 +174,6 @@ bl-table {
 
             .sort-icon {
                 margin-left: 3px;
-                color: $main-background;
             }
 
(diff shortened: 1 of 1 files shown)


dpwatrous:
Is the new behavior that the color doesn't change when you focus the column header?

cRui861:
@dpwatrous Yes, the previous behavior made the gray color of the arrows lighter when hovering over the column names so there wasn't as great of a contrast: 
image

dpwatrous (approved):

dpwatrous:
@cRui861 Gotcha - looks good to me, then!

cRui861:
Included in #2358

PR #2354, opened 2021-07-29
Fix: Unable to log in with personal (MSA) accounts

### LINKED ISSUE #2344: Unable to login with my personal account.
I am trying to log in with my personal account. Which is a @live.com email. But, i keep getting the following error - 
"You can't sign in here with a personal account. Use your work or school account instead." How do i log in with  my personal account?

dpwatrous:
Fixes #2344

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -183,29 +183,19 @@
       }
     },
     "@azure/msal-common": {
-      "version": "4.1.1",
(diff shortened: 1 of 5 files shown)


gingi (approved):
Especially like the unhardcoding of `"common"`, which would have made the change work when I tried it. 😉

PR #2356, opened 2021-07-30
Fix: Resize error name wasn't being displayed properly

dpwatrous:
The ResizeError details object has a format of `{ name: string, value: string }`, not `{ key: string, value: string }`, which the PoolErrorDisplayComponent was using.


gingi (approved):

cRui861:
Fixed #2361

PR #2378, opened 2021-08-20
Caches user auth in secure data store

gingi:
Caches user auth in secure data store

Introduces a persistent cache using the secure data store to store accounts and tokens, which avoids making users log in every time they start the app.

Fixes [AB#10633343](https://msazure.visualstudio.com/AzureBatch/_workitems/edit/10633343/)

### DIFF
--- a/src/client/core/aad/aad-constants.ts
+++ b/src/client/core/aad/aad-constants.ts
@@ -1,4 +1,10 @@
-export const defaultTenant = "organizations";
+export enum TenantPlaceholders {
+    common = "common",
+    organizations = "organizations",
+    consumers = "consumers"
+}
+
+export const defaultTenant = TenantPlaceholders.organizations;
 
 export interface AuthorizeUrlParams {
     response_type: string;
--- a/src/client/core/aad/auth-provider.ts
+++ b/src/client/core/aad/auth-provider.ts
@@ -4,9 +4,12 @@ import {
     ClientApplication,
     PublicClientApplication
 } from "@azure/msal-node";
+import { log } from "@batch-flask/utils";
 import { BatchExplorerApplication } from "..";
+import { SecureDataStore } from "../secure-data-store";
 import { AADConfig } from "./aad-config";
-import { defaultTenant } from "./aad-constants";
+import { defaultTenant, TenantPlaceholders } from "./aad-constants";
+import MSALCachePlugin from "./msal-cache-plugin";
 
 const MSAL_SCOPES = ["user_impersonation"];
 
@@ -18,11 +21,17 @@ export type AuthorizationResult = AuthenticationResult;
 export default class AuthProvider {
     private _clients: StringMap<ClientApplication> = {};
     private _accounts: StringMap<AccountInfo> = {};
+    private _cachePlugin: MSALCachePlugin;
+    private _logoutPromise?: Promise<void>;
+    private _primaryClient?: ClientApplication;
 
     constructor(
         protected app: BatchExplorerApplication,
         protected config: AADConfig
-    ) {}
+    ) {
+        this._cachePlugin =
+            new MSALCachePlugin(app.injector.get(SecureDataStore));
+    }
 
     /**
      * Retrieves an access token
@@ -39,6 +48,10 @@ export default class AuthProvider {
     }): Promise<AuthorizationResult> {
         const { resourceURI, tenantId = defaultTenant, authCodeCallback } = options;
 
+        if (this._logoutPromise) {
+            await this._logoutPromise;
+        }
+
         /**
          * KLUDGE: msal.js does not handle well access tokens across multiple
          * tenants within the same cache. It lets you specify a different
@@ -49,12 +62,15 @@ export default class AuthProvider {
         const client = this._getClient(tenantId);
 
         const authRequest = this._authRequest(resourceURI, tenantId);
-        if (this._accounts[tenantId]) {
+        try {
+            log.debug(`Trying to silently acquire token for '${tenantId}'`);
+            const account = await this._getAccount(tenantId);
             const result = await client.acquireTokenSilent({
-                ...authRequest, account: this._accounts[tenantId]
+                ...authRequest, account
             });
             return result;
-        } else {
+        } catch (silentTokenException) {
+            log.debug(`Trying silent auth code flow (${silentTokenException})`);
             let url, code;
 
             try {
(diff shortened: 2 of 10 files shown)


dpwatrous (changes requested):
Looks like the private NPM registry snuck in here

src/client/core/aad/aad-constants.ts
```diff
@@ -1,5 +1,11 @@
 export const defaultTenant = "organizations";
```
dpwatrous, line 1: Switch to use TenantPlaceholders

src/client/core/aad/auth-provider.ts
```diff
@@ -49,12 +61,16 @@ export default class AuthProvider {
...
+            if (!account) {
+                throw "no account";
+            }
             const result = await client.acquireTokenSilent({
-                ...authRequest, account: this._accounts[tenantId]
+                ...authRequest, account
             });
             return result;
-        } else {
+        } catch (silentTokenException) {
```
dpwatrous, line 73: Should we log something here?

src/client/core/aad/auth-provider.ts
```diff
@@ -63,45 +79,80 @@ export default class AuthProvider {
                     { ...authRequest, prompt: "none" }
                 );
                 code = await authCodeCallback(url, true);
-            } catch (e) {
+            } catch (silentAuthExcpetion) {
```
dpwatrous, line 82: Again, seems like this may be a good place to adding logging so we can troubleshoot auth issues in the future. Also minor typo on silentAuthException

src/client/core/aad/auth-provider.ts
```diff
@@ -63,45 +79,80 @@ export default class AuthProvider {
...
-                this._accounts[tenantId] = result.account;
-            }
             return result;
         }
     }
 
-    public logout(): void {
-        this._removeAccount();
+    public async logout(): Promise<void> {
+        this._logoutPromise = new Promise(resolve => {
```
dpwatrous, line 94: No need to create a new Promise object here. Also, this never actually awaits or returns the logout promise.

Suggestion:

```
public async logout(): Promise<void> {
    this._logoutPromise = this._removeAccounts();
    return this._logoutPromise;
}
```

dpwatrous: (or await this._logoutPromise for that last line - either way)

src/client/core/aad/auth-provider.ts
```diff
@@ -18,11 +20,17 @@ export type AuthorizationResult = AuthenticationResult;
 export default class AuthProvider {
     private _clients: StringMap<ClientApplication> = {};
     private _accounts: StringMap<AccountInfo> = {};
+    private _cachePlugin: MSALCachePlugin;
+    private _logoutPromise: Promise<void>;
```
dpwatrous, line 24: `_logoutPromise` and `_primaryClient` should be marked as optional since `undefined` is a valid value for each of them

src/client/core/aad/auth-provider.ts
```diff
@@ -63,45 +79,80 @@ export default class AuthProvider {
...
+        return client;
     }
 
-    private async _removeAccount(): Promise<void> {
-        for (const tenantId in this._clients) {
-            if (this._accounts[tenantId]) {
-                const cache = this._clients[tenantId].getTokenCache();
-                cache.removeAccount(this._accounts[tenantId]);
+    private async _removeAccounts(): Promise<void> {
+        const cache = this._primaryClient?.getTokenCache();
```
dpwatrous, line 121: This can be undefined. Not sure why Typescript isn't complaining here (maybe we're using permission compiler flags)

src/client/core/aad/aad-constants.ts
gingi: Done.

src/client/core/aad/auth-provider.ts
```diff
@@ -63,45 +79,80 @@ export default class AuthProvider {
...
+        for (const account of accounts) {
+            if (account.tenantId === tenantId) {
+                return this._accounts[tenantId] = account;
+            } else if (!homeAccountId) {
+                homeAccountId = account.homeAccountId;
             }
-            delete this._accounts[tenantId];
         }
+
+        /* SPECIAL CASE: If the tenant is one of the tenant placeholders (e.g.,
```
dpwatrous, line 146: Maybe another good place to put some logging?

gingi: Does this really need to be logged? In 99% of the cases, users will be authenticated against a tenant placeholder, so this is a standard code path. It's only a special case in the sense that we're returning the account for the tenant placeholder using its home account ID as opposed to its resolved ID.

dpwatrous: Good point - I'm still a little concerned about edge cases here. The return type should also be `Promise<AccountInfo | null>` since we're returning null at the end, and we should make sure other code is handling that null. (If that case will never happen, then it should throw an exception instead).

src/client/core/aad/auth-provider.ts
```diff
@@ -49,12 +62,18 @@ export default class AuthProvider {
         const client = this._getClient(tenantId);
 
         const authRequest = this._authRequest(resourceURI, tenantId);
-        if (this._accounts[tenantId]) {
+        try {
+            log.debug(`Trying to silently acquire token for '${tenantId}'`);
+            const account = await this._getAccount(tenantId);
+            if (!account) {
+                throw "no account";
```
dpwatrous, line 69: Let's make this error message more descriptive. Off the top of my head:

`Unable to find a valid AAD account for tenant ${tenantId}`

Also, since we're just throwing an error here anyway, let's switch this._getAccount() to throwing instead of returning null. No need to return null if we're not going to handle it gracefully.

src/client/core/aad/auth-provider.ts
dpwatrous: Update: See my other comment on usage of this method. I think we should just throw an error if there's no better graceful way to handle a missing account.

dpwatrous (changes requested):

src/client/core/aad/auth-provider.ts
gingi: Good point about having `_getAccount()` throw. I just wanted auth to proceed the same way (with auth code flow) whether no account is found or the account is invalid (thrown by MSAL when no valid tokens are associated with the account).

src/client/core/aad/auth-provider.ts
```diff
@@ -63,45 +79,83 @@ export default class AuthProvider {
...
+        /* SPECIAL CASE: If the tenant is one of the tenant placeholders (e.g.,
+         * "common"), fallback to the home tenant account, since the tenant in
+         * the account dictionary are always resolved IDs.
+         */
+        if (tenantId in TenantPlaceholders) {
+            return this._accounts[tenantId] =
+                await cache.getAccountByHomeId(homeAccountId);
+        }
+
+        throw `Unable to find a valid AAD account for tenant ${tenantId}`;
```
dpwatrous, line 158: Just for consistency let's throw an error rather than a string (makes it easier to not have to do type checking when catching errors)

dpwatrous (approved):

cRui861:
Fixes #2412

PR #2381, opened 2021-08-30
Primary download button based on OS

gingi:
Updates the public site to auto-detect the user's OS and set the primary download button to the appropriate binary.


cRui861 (approved):

index.js
```diff
@@ -12,6 +12,12 @@ const feedUrls = {
 
 const feedUrl = feedUrls[buildType];
 
+const OS = {
```
dpwatrous, line 15: *Disclaimer: Complete nitpick, feel free to ignore*

This should probably be a Typescript enum:

```
enum OS = {
    Windows,
    Linux,
    Mac
}
```

Though personally I'm not a big fan of numeric enums unless needed to conserve memory or network bandwidth (it just makes debugging harder), so I'd probably go with this instead:

```
enum OS = {
    Windows = "windows",
    Linux  = "linux",
    Mac  = "mac"
}
```

dpwatrous (approved):

cRui861:
Fixes #2413

PR #2393, opened 2021-10-26
Fix favorite item pinned visual focus indicator through keyboard

cRui861:
Fix bug [AB#574](https://azurebatch.visualstudio.com/3426cbfe-4c9a-4da4-88df-70f025a77017/_workitems/edit/574)

Previously, individual items could not be accessed through keyboard tabbing, only the Delete button for those items could. 
image

Individual items in the favorite item drop down can be accessed through keyboard tabbing now.
image


gingi (approved):

cRui861:
Fixes #2415

PR #2427, opened 2022-02-04
CyberEO Compliance

gingi:
This PR makes BatchExplorer compliant with the [Cybersecurity Executive Order](https://www.microsoft.com/en-us/federal/CyberEO.aspx) (CyberEO):

* Migrates the retail Azure Pipelines agents to secure and compliant 1ES Hosted Pools
* Adds a build task that generates SBOM (Software Bill of Materials) manifests for each package


cRui861:
Fixes #2457

PR #2428, opened 2022-02-04
Merge stable into master following v2.12.0 release 

cRui861:


cRui861:
No updates from stable into master.

PR #2443, opened 2022-02-11
Uses private feeds for NPM and PyPi dependencies

gingi:
For CyberEO compliance, build and distribution pipeline now uses a private feed to manage public dependencies for both NPM and PyPi


scripts/azpipelines/setup-python.sh
```diff
@@ -0,0 +1,34 @@
...
+    export PATH="$PYENV_ROOT/bin:$PATH"
+    pyenv --version
+
+    echo "Installing Python $python_version..."
+    env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install $python_version
+    pyenv global 3.8.5
+    eval "$(pyenv init -)"
+
+    echo "Setting up Python virtual environment..."
+    python -m venv "$pyroot"
```
dpwatrous, line 29: Doesn't this create a virtual environment in the root of the git repo as opposed to inside a .venv directory?

dpwatrous: Also seems like this virtual environment creation/activation should happen outside of the Windows check so we can consistently use a venv for all OSes.

.vsts/python-setup.yml
```diff
@@ -0,0 +1,23 @@
...
+      set -e
+      . "$(Agent.WorkFolder)/.venv/batchexplorer/bin/activate"
+      pip install -r python/requirements.txt
+    condition: ne( variables['Agent.OS'], 'Windows_NT' )
+    displayName: Install Python dependencies
+
+  - powershell: |
+      . .vsts/win/exec.ps1
+      $ErrorActionPreference = "Stop"
+      exec { pip install -r python/requirements.txt }
```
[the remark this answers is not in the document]
gingi: The Windows build never used virtualenv before. I'm not even sure what it accomplishes in Linux/Mac given that we're installing Python as part of the build.

scripts/azpipelines/setup-python.sh
gingi: `$pyroot` is set above (see like 9) to be `.venv/batchexplorer`. As per my other comment, we don't use virtualenv for Windows. I'm not sure why.

.vsts/python-setup.yml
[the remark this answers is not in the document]
dpwatrous: True, I'm not sure it really buys us anything when using pyenv. I think we should be using it on Windows, though, rather than relying on whatever system Python libraries happen to be installed. Given that I still think it makes sense to use one for uniformity across the builds. If we remove pyenv later (which seems totally reasonable) it would also protect us there.

scripts/azpipelines/setup-python.sh
dpwatrous: I'm doing a text search, so I'm not sure how I'm missing it. All I see is this line:

```
pyroot="$AGENT_WORKFOLDER"
```

.vsts/python-setup.yml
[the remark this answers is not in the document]
gingi: Virtualenv on Windows seems to work just fine. PR updated!

scripts/azpipelines/setup-python.sh
dpwatrous: Ah, just noticed it's fixed now - resolving

gingi: It's now set (unconditionally) on [Line 5](https://github.com/Azure/BatchExplorer/blob/9f9789ed06ac2524debd4da7a6d240a00ad01d30/scripts/azpipelines/setup-python.sh#L5)

.vsts/python-setup.yml
```diff
@@ -0,0 +1,29 @@
...
+    displayName: 'Pip Authenticate'
+    inputs:
+      artifactFeeds: 'BatchExplorer'
+
+  - script: |
+      set -e
+      . "$(Agent.WorkFolder)/.venv/batchexplorer/bin/activate"
+      pip install -r python/requirements.txt
+    condition: ne( variables['Agent.OS'], 'Windows_NT' )
+    displayName: Install Python dependencies
```
dpwatrous, line 21: Super minor, but both of these always show up in the list of steps, so we may want to name them something like "Install Python dependencies (Windows)" & "Install Python dependencies (Linux)"

dpwatrous (dismissed):
Good stuff - approved! My other comment's definitely not a blocker so up to you on that one.

.vsts/python-setup.yml
gingi: Noticed that. Makes sense. Updated.

dpwatrous (approved):

cRui861:
Fixes #2458

PR #2444, opened 2022-02-11
Switch from Spectron to Playwright for e2e tests

dpwatrous:


dpwatrous:
This is currently just an experimental branch. It seems to work fairly well locally with some caveats, like not being able to test the final built binary.

dpwatrous:
Abandoning in favor of #2445 which includes Electron and Angular updates.

cRui861:
Fixes #2459

PR #2445, opened 2022-02-12
Update dependencies, switch to Playwright

dpwatrous:
- Updated to Electron 12, Angular 10, Typescript 4
- Replaced Spectron with Playwright
- Updated various other dependencies


dpwatrous:
Looks like we've finally gotten to the point where we can enable Ivy (https://docs.angular.lat/guide/ivy) - added a new commit to do just that. Last time I tried I hit a lot of issues, this time it worked great!

gingi (approved):
Approved, with a suggestion to either replace some Angular Material dependencies or adding a comment about the need to import an unadvertised class name.

gingi (dismissed):

gingi (approved):

PR #2452, opened 2022-02-23
[Fix] Update heatmap node height calculations

cRui861:
- Update heatmap node height calculations for `requiredSlots`
- Update heatmap node tooltip display to include `taskSlotsCount / taskSlotsPerNode`
- Naming update from `task` to `taskSlots`

### DIFF
--- a/src/app/components/pool/graphs/heatmap/nodes-heatmap.component.spec.ts
+++ b/src/app/components/pool/graphs/heatmap/nodes-heatmap.component.spec.ts
@@ -16,6 +16,8 @@ import { NodesHeatmapComponent, NodesHeatmapLegendComponent } from ".";
 
 const legendWidth = 160;
 const svgMargin = 4; // 2x2
+const defaultRunningTasksCount = 2;
+const defaultRunningTaskSlotsCount = 2;
 
 @Component({
     template: `
@@ -135,7 +137,9 @@ describe("NodesHeatmapComponent", () => {
     });
 
     it("should have title with number of tasks running on node", () => {
+        const testTaskSlotsPerNode = 2;
         testComponent.nodes = createNodes(2);
+        testComponent.pool = new Pool({ id: "pool-2", taskSlotsPerNode: testTaskSlotsPerNode });
         fixture.detectChanges();
         const tiles = svg.selectAll("g.node-group");
         tiles.each((d, i, groups) => {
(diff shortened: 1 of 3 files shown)


src/app/components/pool/graphs/heatmap/nodes-heatmap.component.spec.ts
```diff
@@ -144,7 +144,7 @@ describe("NodesHeatmapComponent", () => {
             const title = group.select("title");
 
             expect(title).not.toBeFalsy("Should have a rect in bg");
-            expect(title.text()).toContain("2 tasks running on node");
+            expect(title.text()).toContain(`2 tasks (0/1 slots) running on node (node-${i + 1})`);
```
dpwatrous, line 147: These assertions don't look right - tasks should always consume at least 1 slot.

src/app/components/pool/graphs/heatmap/nodes-heatmap.component.ts
```diff
@@ -341,31 +343,36 @@ export class NodesHeatmapComponent implements AfterViewInit, OnChanges, OnDestro
...
        });
     }
 
-    private _getTaskHeight(tileSize: number, node: Node) {
+    private _getTaskSlotsHeight(tileSize: number, node: Node) {
         const taskSlotsPerNode = this.pool.taskSlotsPerNode;
-        const taskHeight = Math.floor(tileSize / taskSlotsPerNode);
+        const taskSlotsCount = node.runningTaskSlotsCount;
+        let taskSlotsHeight = 0;
+        if (taskSlotsPerNode) {
```
dpwatrous, line 364: Do we need this check? I think taskSlotsPerNode will always be defined, but if that's not the case we can default it to 1 since that's the default at the API level.

cRui861: Yeah I assumed taskSlotsPerNode would always be defined too but this check was supposed to be an extra safety check.

cRui861: Removed the extra check

bgklein:
Not really a comment for this fix, but as user I would want a hover over displaying both slots and number of tasks on the VM. More of a feature there though

cRui861:
Fixes #2461

PR #2469, opened 2022-03-04
February a11y Bug Fixes

cRui861:
February accessibility bug fixes include several improvements on luminosity contrast ratios for buttons and icons. Also the back/forward button is now accessible through keyboard navigation (before these buttons were skipped on keyboard navigation and only accessible through mouse clicks).


dpwatrous (approved):

cRui861:
Fixes #2471

PR #2483, opened 2022-03-17
Bump version to 2.14.0

cRui861:


dpwatrous (changes requested):
This looks like it's reverting the VM SKU API change on master

dpwatrous (approved):

PR #2498, opened 2022-04-28
TS CLI with configure and link commands

gingi:
* Rewrites Batch CLI (to-be-named) in TypeScript (using `yargs`). Now includes help info.
* Adds commands:
     - `configure`: Creates or updates user-config for Batch-related paths
     - `link`: Creates NPM links from Portal extension to local Shared Lib packages
     - `unlink`: Replaces local packages in Portal extension with remote dependencies

The script is still not completely portable since it requires that the user install `ts-node` globally, but otherwise it should work.


util/build-util/tsconfig.json
```diff
@@ -0,0 +1,43 @@
+{
```
dpwatrous, line 1: Can we reference the base tsconfig here or does that break using it standalone?

util/build-util/tsconfig.json
```diff
@@ -0,0 +1,43 @@
...
+    "exactOptionalPropertyTypes": true,
+    "experimentalDecorators": true,
+    "forceConsistentCasingInFileNames": true,
+    "importHelpers": true,
+    "importsNotUsedAsValues": "error",
+    "incremental": true,
+    "lib": [
+      "es2021"
+    ],
+    "module": "UMD",
```
dpwatrous, line 21: Seems like this should be "CommonJS" since we're running this in Node.js only

docs/setup.md
```diff
@@ -70,3 +70,29 @@ Run dev electron (to use the dev server you need to run this)
...
+## Working on the Shared Libraries and the Portal Extension
+
+The Portal Extension relies on packages provided by the shared libraries. The portal installs the necessary dependencies that are published on a remote registry. If you want to work on the portal in tandem with local, unpublished versions of the shared libraries, you will need to run the following commands:
+
+
+1. Install the Batch Utility CLI
+
+```shell
+cd ./util/build-util
+yarn global add $PWD
```
dpwatrous, line 83: Is $PWD needed here? Does `yarn global add .` work? I believe $PWD won't work in a Windows command prompt (but is fine in Powershell).

gingi: Yeah, I thought of that too, but no, `yarn global add .` alas doesn't work. :(

util/build-util/tsconfig.json
gingi: Fixed.

docs/setup.md
gingi: Fixed documentation.

docs/setup.md
```diff
@@ -29,44 +39,68 @@ Please also take a look at the [coding guidelines](coding-guidelines.md) for thi
...
+* EditorConfig for VS Code (Important)
+* ESLint
 * Debugger for Chrome
 
 ## Detailed commands
 
 **Magic command** (starts the server and electron in dev mode)
 
-```shell
+```bash
```
dpwatrous, line 50: Does GitHub not support "shell"? These aren't bash-specific commands

packages/react/src/ui-react/components/__tests__/button.spec.tsx
```diff
@@ -9,9 +9,16 @@ describe("Button component", () => {
     beforeEach(() => initMockBrowserEnvironment());
 
     test("Simple button", async () => {
-        const { container } = render(<Button label="I am a button" />);
+        const { container } = render(<Button />);
```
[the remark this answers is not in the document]
gingi: The unit test ensured that this is the only a11y violation when no label is supplied, but it is extraneous. Reverting.

util/build-util/util.ts
```diff
@@ -0,0 +1,279 @@
...
+
+export function readJsonOrDefault(filename: string, defaultJson = {}) {
+    if (fs.existsSync(filename)) {
+        return require(filename);
+    }
+    return defaultJson;
+}
+
+export async function saveJson(filename: string, json: unknown) {
+    const indent = await exports.getEditorConfig("indent_size", filename);
```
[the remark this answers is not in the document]
gingi: This uses the .editorconfig that's in the root directory of Batch Explorer, and it definitely exists. But I added a default to fall back on. This routine does not use parse file extensions to determine what format to save it in. So `saveJson("file.config", jsonData)` would work just as well.

PR #2517, opened 2022-06-08
Added checkbox and radio button group form controls

skapur12:
[Checkbox control work item](https://msazure.visualstudio.com/AzureBatch/_workitems/edit/14215432/)
- Added wrapped checkbox form control, unit tests for it, & integrated it into the playground

[Radio button group control work item](https://msazure.visualstudio.com/AzureBatch/_workitems/edit/14215411/)
- Added wrapped radio button form control, unit tests for it, & integrated it into the playground

### DIFF
--- a/packages/playground/src/ui-playground/new-functions.tsx
+++ b/packages/playground/src/ui-playground/new-functions.tsx
@@ -1,0 +1,35 @@
+// For any functions involving shared libraries form controls
+// used throughout the playground
+
+import { RadioButtonOption } from "@batch/ui-react/lib/components/form/radio-button";
+
+/*
+ * Syncs the updated value of a shared libraries text field with a playground component
+ */
+export function _newTextfieldOnChange(
+    setValue: React.Dispatch<React.SetStateAction<string>>
+): (newValue?: string | undefined) => void {
+    function Result(newValue?: string | undefined): void {
+        if (newValue) {
+            setValue(newValue);
+        }
+    }
+    return Result;
+}
+
+/*
+ * Syncs the updated value of a shared libraries radio button group with a playground component
+ */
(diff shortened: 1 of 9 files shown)


packages/playground/src/ui-playground/new-functions.tsx
```diff
@@ -0,0 +1,35 @@
+// For any functions involving shared libraries form controls
+// used throughout the playground
+
+import { RadioButtonOption } from "@batch/ui-react/lib/components/form/radio-button";
+
+/*
+ * Syncs the updated value of a shared libraries text field with a playground component
+ */
+export function _newTextfieldOnChange(
```
gingi, line 9: Is there a reason why these functions are prefixed with an underscore? That's usually the convention for private methods and variables.

packages/react/src/ui-react/components/form/checkbox.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export function Checkbox<V>(props: CheckboxProps<V>): JSX.Element {
+    if (props.hidden) {
+        return <></>;
+    }
+
+    return (
+        <FluentCheckbox
+            id={props.id}
```
gingi, line 22: Just a minor note that instead of passing individual props, you can pass the entire props object with the spread operator, e.g., `<FluentCheckbox ...props onChange={/* ... */} />`.

gingi (approved):
These are great additions.

packages/react/src/ui-react/components/form/checkbox.tsx
skapur12: Thanks! Just changed that for both form controls

gingi (approved):

PR #2560, opened 2022-08-17
HTTP client with authenticated requests, plus misc

gingi:
* Storage Account Service
* Subscription Service
* Basic form onChange support to allow parameters to detect when other parameters change
* Basic skeleton for specifying parameter dependencies (WIP)
* Fixes some linting warnings
* Updates dependencies
* Minor tweaks and cleanup


packages/service/src/ui-service/storage/__tests__/storage-account-service.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+describe("StorageAccountService", () => {
+    let httpClient: MockHttpClient;
+
+    beforeEach(() => {
+        initMockEnvironment();
+        httpClient = getMockEnvironment().getHttpClient();
+    });
+
+    test("list()", async () => {
+        const service = new FakeStorageAccountService();
```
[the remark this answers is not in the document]
dpwatrous: Also I think you found a bug in MockHttpClient. We should throw if a test ends without an expected call being made.

packages/service/src/ui-service/subscription/fake-subscription-service.ts
```diff
@@ -0,0 +1,30 @@
...
+    { id: "/fake/sub1", displayName: "Subscription One" },
+    { id: "/fake/sub2", displayName: "Subscription Two" },
+    { id: "/fake/sub3", displayName: "Subscription Three" },
+    { id: "/fake/sub4", displayName: "Subscription Four" },
+    { id: "/fake/badsub", displayName: "Bad Subscription" },
+];
+
+export class FakeSubscriptionService implements SubscriptionService {
+    public list(): Promise<Subscription[]> {
+        return Promise.resolve(subscriptions as Subscription[]);
```
dpwatrous, line 14: Similar to the other comment, it doesn't really matter a whole lot but IMO `async` with a simple `return` statement is so nice syntactically.

packages/service/src/ui-service/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const Endpoints = {
+    arm: `https://management.azure.com`,
+};
+
+export const ApiVersion = {
```
dpwatrous, line 5: We should be explicit about what API versions these are for, since they're different for every RP.

packages/service/src/ui-service/resource-service.ts
```diff
@@ -0,0 +1,11 @@
+export interface ResourceService<T> {
+    list(subscriptionId?: string): Promise<T[]>;
```
dpwatrous, line 2: Listing by subscription ID may only be possible for top-level ARM resources.

dpwatrous (changes requested):
Let's follow up on the form issues. Definitely some problems there with doing validation at the control level instead of the form or parameter level.

.vscode/settings.json
```diff
@@ -1,63 +1,61 @@
...
   // Per Language Formatting
   // Keep this up to date with .prettierignore
   // Run Prettier first, followed by ESLint
   "[javascript]": {
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": [
-        "source.formatDocument",
-        "source.fixAll.eslint"
+      "source.formatDocument",
+      "source.fixAll"
```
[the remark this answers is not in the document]
gingi: Removed this and the whitespace changes (introduced by a local VSCode extension; sorry!). Also adding VSCode Icons as a recommended extension, since we already have a setting for it (Line 3).

desktop/.stylelintignore
```diff
@@ -0,0 +1,3 @@
+**/*.md
```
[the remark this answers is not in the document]
gingi: Moved to a different branch (feature/quality-improvements).

desktop/package.json
```diff
@@ -160,14 +160,15 @@
     "node-fetch": "^2.6.7",
     "nyc": "^15.1.0",
     "playwright": "^1.18.1",
+    "postcss-scss": "^4.0.4",
```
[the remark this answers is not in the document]
gingi: It's a required dependency of stylelint@14 (moved to a different branch against master)

desktop/src/@batch-flask/ui/breadcrumbs/breadcrumb-group/breadcrumb-group.scss
```diff
@@ -4,7 +4,7 @@ bl-breadcrumb-group {
     $header-margin: 4px;
     $separation: 3px;
     $height: $header-height - 2 * $header-margin;
-    $half-height: $height / 2;
+    $half-height: calc($height / 2);
```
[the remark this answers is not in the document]
gingi: Moved to a different branch (feature/quality-improvements).

packages/service/src/ui-service/storage/__tests__/storage-account-service.spec.ts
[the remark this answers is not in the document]
gingi: Added a `remainingAssertions()` method on `MockHttpClient` to allow suites to check for post-test assertions.

desktop/src/common/constants/constants.ts
```diff
@@ -229,8 +229,6 @@ export const rendererEvents = {
     navigateTo: "navigate-to",
 };
 
-export const isRenderer = (process && process.type === "renderer");
```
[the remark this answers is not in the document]
gingi: It appeared as a linting warning, and then I realized that it's not used anywhere in the codebase. (I moved this to a separate branch targeted against master.) Should I just keep it?

packages/react/src/ui-react/hooks/use-async-effect.ts
```diff
@@ -37,14 +36,9 @@ export function useAsyncEffect(
...
-                // Cleanup functions should be returned
-                return ret;
-            }
-            // Void otherwise
-            return;
+            Promise.resolve(ret);
         },
         // eslint-disable-next-line react-hooks/exhaustive-deps
-        deps
+        deps ?? []
```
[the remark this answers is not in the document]
gingi: Reverted. The other change should obviate the need to fall back to an empty array.

packages/common/src/ui-common/form/form.ts
```diff
@@ -153,6 +154,10 @@ export type ValidationOpts = {
 
 const defaultValidationOpts = { force: false };
 
+export type FormChangeHandler<V> = (newValues: V, oldValues: V) => void;
+
+export type FormEventType = "change";
```
dpwatrous, line 159: The "validate" event is missing here

packages/common/src/ui-common/form/form.ts
```diff
@@ -249,6 +254,27 @@ export interface Form<V extends FormValues> {
...
+     * Subscribe to form events. When the form event is triggered, the handler
+     * is called. This method returns the handler as a convenience to allow for
+     * easy cleanup.
+     * @param eventType The type of event to listen to
+     * @param handler The callback invoked when the event is triggered
+     * @see off()
+     */
+    on(
+        eventType: FormEventType,
+        handler: FormChangeHandler<V>
```
dpwatrous, line 268: Change isn't the only type of handler. We also have a "validate" event

packages/common/src/ui-common/form/form.ts
```diff
@@ -433,6 +459,12 @@ class FormImpl<V extends FormValues> implements Form<V> {
      */
     private _childEntries: OrderedMap<string, Entry<V>> = new OrderedMap();
 
+    /**
+     * Collection for looking up other entries whose value this entry depends
+     * on.
+     */
+    private _entryDependencies: Map<string, string[]> = new Map();
```
dpwatrous, line 466: Should the keys here be `keyof V`?

packages/common/src/ui-common/form/form.ts
```diff
@@ -882,10 +939,10 @@ export class SubForm<
         return this.form.getSubForm(name);
     }
 
-    updateValue<
-        SK extends Extract<keyof S, string>,
-        S2 extends S[SK] & FormValues
```
dpwatrous, line 887: I thought this `& FormValues` was needed to give Typescript a hint that subforms have values that implement the FormValues interface, but maybe not. Frankly subforms aren't fully fleshed out outside of tests, so it's probably fine.

packages/common/src/ui-common/form/form.ts
```diff
@@ -910,6 +967,17 @@ export class SubForm<
     async waitForValidation(): Promise<ValidationStatus | undefined> {
         return this.form.waitForValidation();
     }
+
+    on(
+        eventType: FormEventType,
+        handler: FormChangeHandler<S>
```
dpwatrous, line 973: See the previous comment on the "validate" event

packages/common/src/ui-common/index.ts
```diff
@@ -9,6 +10,15 @@ export {
     EnvironmentMode,
     DependencyName,
 } from "./environment";
+export {
+    AbstractHttpClient,
+    FetchHttpClient,
+    HttpClient,
+    HttpRequestInit,
+    MockHttpClient,
```
dpwatrous, line 18: IMO let's not export mocks/fakes in index.ts. I *think* this still causes issues where webpack can't treeshake out these exports.

packages/react/src/ui-react/components/form/__tests__/parameter-type.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+import * as React from "react";
+import { initMockBrowserEnvironment } from "../../../environment";
+import { runAxe } from "../../../test-util/a11y";
+import {
+    StorageAccountDropdown,
+    SubscriptionDropdown,
+} from "../parameter-type";
+
+/* KLUDGE: the parameter has to be called "subscriptionId" until we can specify
+ * dependencies in parameter types
```
dpwatrous, line 15: Can this use dependencies now?

packages/react/src/ui-react/components/form/parameter-type.tsx
```diff
@@ -157,33 +167,98 @@ export function StringParamTextField<
...
+        } catch (error) {
+            setValidationStatus(new ValidationStatus("error", error + ""));
+        } finally {
+            setStorageAccounts(accounts);
+            setLoading(false);
+        }
+    }, [subscriptionId]);
+
+    useEffect(() => {
+        const handler = form.on("change", (values: FormValues) =>
```
dpwatrous, line 208: I don't think an on change event listener is necessary since the most recent form value is always passed down in the props.

packages/react/src/ui-react/components/form/parameter-type.tsx
```diff
@@ -157,33 +167,98 @@ export function StringParamTextField<
...
-    const [subscriptions, setSubscriptions] = React.useState<
+    useAsyncEffect(async () => {
+        let accounts: StorageAccount[] = [];
+        try {
+            if (subscriptionId) {
+                accounts = await service.list(subscriptionId);
+            }
+            setValidationStatus(null);
+        } catch (error) {
+            setValidationStatus(new ValidationStatus("error", error + ""));
```
dpwatrous, line 200: I'm not sure this should be a validation error, but it is something I haven't thought about. Let's follow up.

packages/react/src/ui-react/hooks/use-async-effect.ts
```diff
@@ -32,16 +32,25 @@ export type AsyncEffectCallback = () => Promise<unknown>;
...
+            // If the returned value is not an async function, just treat it
+            // like useEffect
+            if (typeof ret === "function") {
                 return ret;
             }
-            // Void otherwise
+
+            if (isPromiseLike(ret)) {
+                ret.then((cleanup) => {
+                    if (typeof cleanup === "function") {
```
dpwatrous, line 49: Instead of checking at runtime here, can we change the type of AsyncEffectCallback to get better type checking?

I think this may work:

```
export type AsyncEffectCallback = () =>
    | void
    | (() => void)
    | Promise<void>
    | Promise<() => void>;
```

Or maybe better, we could forget about handling the non-promise case (because why are you using useAsyncEffect if it doesn't return a promise?), and simplify to:

```
export type AsyncEffectCallback = () => Promise<void | (() => void)>;
```

Not sure why I had <unknown> in the type before. Probably no good reason.

dpwatrous: If we do the latter we can also remove the isPromiseLike check since it's always going to be a promise or undefined.

packages/service/src/ui-service/storage/storage-account-service.ts
```diff
@@ -0,0 +1,49 @@
...
+    update(account: StorageAccount): Promise<void>;
+}
+
+export class StorageAccountServiceImpl
+    extends AbstractHttpService
+    implements StorageAccountService
+{
+    public async list(subscriptionId: string): Promise<StorageAccount[]> {
+        const response = await this.httpClient.get(
+            `${Endpoints.arm}/subscriptions/${formatSubscriptionId(
```
dpwatrous, line 20: The call to formatSubscriptionId seems out of place. This function should only accept a valid subscription ID.

packages/service/src/ui-service/storage/storage-account-service.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+    public async remove(): Promise<void> {
+        return;
+    }
+    public async update(): Promise<void> {
+        return;
+    }
+}
+
+function formatSubscriptionId(id: string) {
```
dpwatrous, line 47: Do we need this?

packages/service/src/ui-service/view/resource-models.ts
```diff
@@ -0,0 +1,3 @@
+export interface ResourceListResponse<T> {
```
dpwatrous, line 1: This seems out of place in /view/ since that's currently all about wrapping models in observable views (using MobX). Maybe we should have a top-level arm.ts or something, since this is about arm resources?

dpwatrous: Speaking of which, can we rename this ArmResourceListResponse?

dpwatrous (changes requested):

packages/common/src/ui-common/index.ts
gingi: Makes sense. Should I do the same with `getMockEnvironment` and `initMockEnvironment` at [the top of the file](https://github.com/Azure/BatchExplorer/blob/vnext/packages/common/src/ui-common/index.ts#L4)?

packages/common/src/ui-common/environment/environment.ts
```diff
@@ -68,6 +73,8 @@ export enum DependencyName {
     Logger = "logger",
     Localizer = "localizer",
     HttpClient = "httpClient",
+    StorageAccountService = "storageAccount",
```
dpwatrous, line 76: This should move up to the service package rather than be part of the base environment in common, since services aren't a concept in common.

packages/common/src/ui-common/form/form.ts
```diff
@@ -153,6 +153,10 @@ export type ValidationOpts = {
 
 const defaultValidationOpts = { force: false };
 
+export type FormEventHandler<V> = (newValues: V, oldValues: V) => void;
```
dpwatrous, line 156: This doesn't look quite right - it assumes all form event handlers will have these 2 arguments. That's not really the case for "validate" for example.

packages/service/src/ui-service/arm.ts
```diff
@@ -0,0 +1,3 @@
+export interface ArmResourceListResponse<T> {
+    value: T[];
```
dpwatrous, line 2: Would be good to add `nextLink?: string` here as well.

dpwatrous (changes requested):

packages/common/src/ui-common/form/form.ts
dpwatrous: I'm fine with handling this in a later PR, FYI

dpwatrous (approved):

PR #2569, opened 2022-09-01
Add basic shared libraries localization interface

skapur12:
Implemented a basic localization interface for the shared libraries that the portal and Batch Explorer can use

### DIFF
--- a/packages/common/src/ui-common/action/action.ts
+++ b/packages/common/src/ui-common/action/action.ts
@@ -25,16 +25,24 @@ export type ActionExecutionResult = {
 export abstract class AbstractAction<V extends FormValues>
     implements Action<V>
 {
-    form: Form<V>;
+    private _initialValues: V;
+    private _form?: Form<V>;
 
+    get form(): Form<V> {
+        // Lazily create the form the first time it's accessed
+        if (!this._form) {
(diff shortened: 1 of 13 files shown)


packages/common/src/ui-common/action/action.ts
```diff
@@ -25,16 +25,24 @@ export type ActionExecutionResult = {
...
+            this._form = this.buildForm(this._initialValues);
+            this.form.onValidateSync = (snapshot) => {
+                return this._validateSync(snapshot);
+            };
+            this.form.onValidateAsync = (snapshot) => {
+                return this._validateAsync(snapshot);
+            };
+        }
+        return this._form;
+    }
```
dpwatrous, line 43: Nit: Add a space between the form accessor and the constructor function.

packages/common/src/ui-common/localization/mock-localizer.ts
```diff
@@ -0,0 +1,16 @@
+import { Localizer } from "./localizer";
+
+export class MockLocalizer implements Localizer {
```
dpwatrous, line 3: Let's call this something like StandardLocalizer instead of MockLocalizer. This isn't really a mock implementation

dpwatrous (approved):
Looks good!

PR #2572, opened 2022-09-08
No longer need to ignore a11y rules for FluentUI

dpwatrous:


gingi (approved):

dpwatrous:
Not sure what I was looking at before, but it appears these ignored rules are still needed.

PR #2574, opened 2022-09-16
Add batchSupportEndOfLife to show image deprecation warnings

cRui861:
- Deprecation warnings will show for images that have an end of life date (batchSupportEndOfLife)
- "Learn more" link will not show for images that don't have an associated link for users to get more information on the deprecation. Warning with date of deprecation will still appear for users.
- Add deprecation link for WindowsServer "with-container" images

Example during pool creation:
image

Example for pool details:
image


src/app/components/pool/action/add/pool-create-basic-dialog.html
```diff
@@ -29,18 +29,19 @@
...
+                    </div>
+                </bl-banner>
+            </ng-container>
             <ng-container *ngIf="cloudServiceConfiguration">
-                <bl-banner class="cloud-services-deprecation-warning" type="warning">
+                <bl-banner class="create-pool-deprecation-warning" type="warning">
                     <div message>
                         Cloud Services pools are deprecated and will be retired on 29 February 2024. Existing ‘CloudServiceConfiguration’ pools will need to be deleted and new
-                        <a class="cloud-services-warning-alternative-link" (click)="openLink('https://docs.microsoft.com/en-us/azure/batch/batch-pool-cloud-service-to-virtual-machine-configuration?WT.mc_id=Portal-Microsoft_Azure_Batch')" href="javascript:void(0)">‘VirtualMachineConfiguration’ pools created. </a>
+                        <a class="create-pool-warning-alternative-link" (click)="openLink('https://docs.microsoft.com/en-us/azure/batch/batch-pool-cloud-service-to-virtual-machine-configuration?WT.mc_id=Portal-Microsoft_Azure_Batch')" href="javascript:void(0)">‘VirtualMachineConfiguration’ pools created. </a>
```
dpwatrous, line 44: I realize this isn't actually new to this PR, but when linking to documentation, omit the locale (en-us in this case). Should be: https://docs.microsoft.com/azure/batch/batch-pool-cloud-service-to-virtual-machine-configuration?WT.mc_id=Portal-Microsoft_Azure_Batch

src/common/constants/pool-constants.ts
```diff
@@ -0,0 +1,17 @@
+export const DeprecatedContainerImages = [
```
dpwatrous, line 1: Can this go as a utility function in `util/pool-utils.ts`?

Something like:

isDeprecatedContainerImage: (imageId: string): boolean {
    // Check imageId.toLowerCase() is in the list of deprecated images
}

src/app/components/pool/action/add/pool-create-basic-dialog.component.ts
```diff
@@ -38,6 +39,14 @@ export class PoolCreateBasicDialogComponent extends DynamicForm<Pool, PoolCreate
         return this._osControl.value && this._osControl.value.cloudServiceConfiguration;
     }
 
+    public get isSelectedContainerImageDeprecated() {
+        if (this._osControl.value && this._osControl.value.virtualMachineConfiguration) {
+            const selectedImage = this._osControl.value.virtualMachineConfiguration.imageReference.sku;
+            return DeprecatedContainerImages.includes(selectedImage);
```
dpwatrous, line 45: I'm not sure it matters (can't remember if pool properties are case-preserving), but we should probably do a case-insensitive comparison.

dpwatrous (changes requested):
Looks good overall, just a few minor suggestions.

src/common/constants/pool-constants.ts
cRui861: Yeah that's a good call, I wasn't sure where to put that list of deprecated container images at first which is why I made a new constants file. I changed it to your suggestion now.

src/app/components/pool/action/add/pool-create-basic-dialog.component.ts
cRui861: Thanks for catching that!

dpwatrous:
On the pool overview, can we tweak the message slightly? Something like:

"This pool's image is nearing its end-of-life date of $DATE. After this date, new nodes may not function properly."

(This could probably use some more word-smithing)

cRui861:
> On the pool overview, can we tweak the message slightly? Something like:
> 
> "This pool's image is nearing its end-of-life date of $DATE. After this date, new nodes may not function properly."
> 
> (This could probably use some more word-smithing)

Kept it mostly the same but moved the actual date to the second sentence:
"This pool's image is nearing its end-of-life date. After $DATE, new nodes may not function properly."

src/app/components/pool/action/add/pool-create-basic-dialog.component.ts
```diff
@@ -38,6 +39,30 @@ export class PoolCreateBasicDialogComponent extends DynamicForm<Pool, PoolCreate
         return this._osControl.value && this._osControl.value.cloudServiceConfiguration;
     }
 
+    public get isSelectedImageDeprecated() {
+        if (this._osControl.value && this._osControl.value.virtualMachineConfiguration) {
```
dpwatrous, line 43: FYI, Typescript has this great syntax now, where it will return undefined if `value` is undefined instead of erroring out:

```
const config = this._osControl.value?.virtualMachineConfiguration;
if (config) {
   // ...etc
}
```

dpwatrous: (No need to change this unless you want)

src/app/components/pool/action/add/pool-create-basic-dialog.html
```diff
@@ -29,18 +29,19 @@
...
-                <div message>
-                    The Windows container images below will be removed from marketplace starting on September 22, 2022. After this date, new pools using it may not be created from the Azure Batch Explorer.
-                    <a (click)="openLink('https://github.com/Azure/Batch/issues/136')" href="javascript:void(0)">Click here </a>
-                    for more information.
-                </div>
-            </bl-banner>
+            <ng-container *ngIf="isSelectedImageDeprecated">
+                <bl-banner class="create-pool-deprecation-warning" type="warning">
+                    <div message>
+                        The selected image is nearing its end of life date on {{selectedVirtualMachineImageEndOfLifeDate}}. API calls to create new pools using this image may continue to be supported for up to 60 days after the end of life date.
```
dpwatrous, line 35: Can we pull this from an i18n.yml file?

src/app/utils/pool-utils.ts
```diff
@@ -174,6 +174,14 @@ export class PoolUtils {
         return "cloud";
     }
 
+    public static getEndOfLifeHyperlink(imageName: string): string {
+        if (imageName.toLowerCase().includes("with-containers")) {
```
dpwatrous, line 178: Let's make a dictionary and match on the exact string rather than doing substring matching here. I'm not sure this is a problem right now since the replacement image is called "windows_with_mirantis_container_runtime_2019" but there's no reason another image couldn't have "with-containers" in it.

src/app/utils/pool-utils.ts
```diff
@@ -174,6 +174,14 @@ export class PoolUtils {
...
+            return 'https://github.com/Azure/Batch/issues/136';
+        } else {
+            return "";
```
dpwatrous, line 181: This is largely opinion, but IMO it's clearer if this function returns null or undefined when it doesn't find a link. That way Typescript will make consuming code handle the possibility of a null/undefined value instead of having to know to handle empty string (which Typescript can't help with).

src/app/components/pool/action/add/pool-create-basic-dialog.component.ts
```diff
@@ -38,6 +39,30 @@ export class PoolCreateBasicDialogComponent extends DynamicForm<Pool, PoolCreate
...
+            const config = this._osControl.value.virtualMachineConfiguration;
+            const isDeprecatedImage = config && config.batchSupportEndOfLife;
+            if (isDeprecatedImage) {
```
dpwatrous, line 46: Instead of just checking for the presence of an end of life date, I think we should see how close it is and update the message accordingly. Let's just match the logic in the portal for this. I don't think it would be out of the question to have an end of life date defined that's 2 years in the future, but it wouldn't really be accurate to say it's "nearing" the end of life date in that case.

dpwatrous (changes requested):

src/app/components/pool/action/add/pool-create-basic-dialog.component.ts
cRui861: Updated to the suggested syntax

src/app/components/pool/action/add/pool-create-basic-dialog.html
cRui861: Yes, thanks for catching that.

src/app/utils/pool-utils.ts
cRui861: Added the suggested dictionary. I know that Typescript has a Set() but on average it seems to still be less performant than this dictionary. I know that the values aren't very useful though. Let me know if there is a more elegant solution.

src/app/utils/pool-utils.ts
cRui861: Updated

src/app/utils/pool-utils.ts
cRui861: Updated this to just an array since performance-wise it wouldn't be that different.

src/app/components/pool/action/add/pool-create-basic-dialog.component.ts
cRui861: Added new EOL messages to match those that are used in the Portal.

dpwatrous (approved):

cRui861:
Fixed #2635

PR #2575, opened 2022-09-28
fix: multiple accessibility issues

wanghoppe:
AB#608
AB#609
AB#674
AB#861

### DIFF
--- a/src/app/components/pool/action/add/os-picker/os-offer-tile/os-offer-tile.html
+++ b/src/app/components/pool/action/add/os-picker/os-offer-tile/os-offer-tile.html
@@ -16,7 +16,7 @@
     class="select-version"
     [matMenuTriggerFor]="menu"
     matTooltip="{{'os-offer-tile.select-version-button.tooltip' | i18n}}"
-    title="{{'os-offer-tile.select-version-button.tooltip' | i18n}}">
+    title="{{'os-offer-tile.select-version-button.tooltip' | i18n}} {{name}}">
     <i class="fa fa-cog"></i>
 </button>
 <mat-menu #menu="matMenu">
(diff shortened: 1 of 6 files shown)


src/app/components/settings/settings.html
```diff
@@ -1,5 +1,5 @@
 <div class="settings-header">
-    <h1 tabindex="0" autofocus>{{'settings.title' | i18n}}</h1>
+    <h1>{{'settings.title' | i18n}}</h1>
```
dpwatrous, line 2: It looks like we have this same behavior in auth-settings.html. I'm guessing it was just copy/pasted from settings.html. Can we change that in this PR as well?

src/app/components/pool/action/add/os-picker/os-offer-tile/os-offer-tile.html
```diff
@@ -16,7 +16,7 @@
     class="select-version"
     [matMenuTriggerFor]="menu"
     matTooltip="{{'os-offer-tile.select-version-button.tooltip' | i18n}}"
-    title="{{'os-offer-tile.select-version-button.tooltip' | i18n}}">
+    title="{{'os-offer-tile.select-version-button.tooltip' | i18n}} {{name}}">
```
dpwatrous, line 19: A couple of minor suggestions:

- Let's use `prettyName `instead of `name `so we get "almalinux hpc" instead of "almalinux-hpc"
- Adding parentheses around the name would make it more grammatically correct. So: "Select version (almalinux hpc)" instead of "Select version almalinux hpc". Another option would be to use a template string, but parentheses work for me.

dpwatrous (approved):

cRui861:
Fixes #2634

PR #2579, opened 2022-10-13
fix: Create button for storage container under data menu is broken

wanghoppe:
Fix: [AB#1011](https://azurebatch.visualstudio.com/3426cbfe-4c9a-4da4-88df-70f025a77017/_workitems/edit/1011),
To repro can see [AB#609](https://azurebatch.visualstudio.com/3426cbfe-4c9a-4da4-88df-70f025a77017/_workitems/edit/609)

### DIFF
--- a/src/@batch-flask/ui/buttons/clickable/clickable.component.ts
+++ b/src/@batch-flask/ui/buttons/clickable/clickable.component.ts
@@ -12,7 +12,6 @@ import {
     Self,
     SimpleChanges,
 } from "@angular/core";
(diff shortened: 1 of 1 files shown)


wanghoppe:
Repro can see #AB609

wanghoppe:
Using <bl-button> and matMenuTriggerFor directive together has the problem and deleting those codes in PR fixes it. It seems that the add button in the data Menu is the only use case, so hopefully, it won't break others

gingi (approved):

cRui861:
Fixed #2636

PR #2580, opened 2022-10-18
Updates to new Azure Storage SDK

gingi:
Migrates from `azure-storage` 2.10.7 to `@azure/storage-blob` 12.11.0.

Other minor changes:

* Fixes poll rate for refresh by EntityView
* Improves the storage container delete dialog

image image


src/client/core/storage/storage-blob.service.ts
```diff
@@ -0,0 +1,359 @@
...
+            }
+            for (const blob of segment.blobItems) {
+                blobs.push({
+                    name: blob.name,
+                    url: `${containerName}/${blob.name}`,
+                    isDirectory: false,
+                    properties: {
+                        contentLength: blob.properties.contentLength,
+                        contentType: blob.properties.contentType,
+                        creationTime: null,
```
dpwatrous, line 251: See other comment on creationTime

src/client/core/storage/storage-blob.service.ts
```diff
@@ -0,0 +1,359 @@
...
+                    version: container.version,
+                    deleted: container.deleted,
+                    id: container.name
+                });
+            }
+            cToken = page.continuationToken;
+        }
+        return {
+            data: containers,
+            continuationToken: cToken
```
dpwatrous, line 180: If this did iterate over all of the pages we shouldn't return a continuation token. If something else is consuming this token we shouldn't be iterating over all the pages.

dpwatrous (changes requested):

src/client/core/storage/storage-blob.service.ts
```diff
@@ -0,0 +1,359 @@
...
+        return {
+            data: {
+                name: blobName,
+                url: `${containerName}/${blobPath}`,
+                isDirectory: false,
+                properties: {
+                    contentLength: props.contentLength,
+                    contentType: props.contentType,
+                    creationTime: null,
+                    lastModified: props.lastModified,
```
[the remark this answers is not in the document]
gingi: Good call. It's `createdOn` in the [JS SDK](https://learn.microsoft.com/en-us/javascript/api/@azure/storage-blob/blobproperties?view=azure-node-latest#@azure-storage-blob-blobproperties-createdon)

src/client/core/storage/storage-blob.service.ts
gingi: As far as I can tell, at one point there were limits that were passed in, but they're no longer used. There's a `limit` parameter that was passed in from the `FileNavigator` to control the response page size, but it doesn't restrict how many pages are returned, so it's essentially a performance heuristic.

The `continuationToken` is handled gracefully by our code, such that if you return a partial response with a token, you can resume SDK calls from that point. It doesn't seem to have a rendering benefit, because the list is only rendered when all the items are returned. Maybe if there's some user-managed paging of blobs and containers, this could help, but I don't think we have any components that do that.

Based on some performance measures, I set some newer default page sizes for containers and blobs, and added an optional `maxPages` option, so that in some future time we can use it for paging or infinite scrolling. Also added some unit tests to show this behavior.

Lastly, I thought about handling a hard limit on the number of items, but it's not easy to implement, unless we start caching responses or require some other arguments. The `continuationToken` is opaque and lacks the granularity to resume mid-page. It can only be used to resume at the start of pages. Anyhow, probably not worth worrying about.

dpwatrous (approved):

src/app/services/storage/blob-storage-client-proxy.ts
```diff
@@ -28,370 +21,239 @@ export interface ListBlobResponse {
...
-                            };
-                        })),
-                        continuationToken: result.continuationToken,
-                    });
-                });
-        });
+    public async listBlobs(
+        containerName: string,
+        options: ListBlobOptions,
+        continuationToken?: any
```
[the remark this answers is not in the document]
dpwatrous: This isn't really a new issue, so up to you if you want to fix it in this PR

dpwatrous (approved):

cRui861:
Fixed #2637

PR #2584, opened 2022-10-31
fix: announce ‘no match’ information on subscription filter

wanghoppe:
fix: [AB#758](https://azurebatch.visualstudio.com/3426cbfe-4c9a-4da4-88df-70f025a77017/_workitems/edit/758)

### DIFF
--- a/src/@batch-flask/ui/select/select-dropdown/select-dropdown.component.ts
+++ b/src/@batch-flask/ui/select/select-dropdown/select-dropdown.component.ts
@@ -1,9 +1,9 @@
 import { LiveAnnouncer } from "@angular/cdk/a11y";
 import {
     ChangeDetectionStrategy, ChangeDetectorRef, Component, ElementRef,
-    HostBinding, Inject, forwardRef,
+    HostBinding, Inject, forwardRef
 } from "@angular/core";
-import { ListKeyNavigator } from "@batch-flask/core";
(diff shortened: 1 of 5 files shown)


src/@batch-flask/ui/select/select-dropdown/select-dropdown.component.ts
```diff
@@ -124,4 +128,13 @@ export class SelectDropdownComponent {
         this.keyNavigator.items = this.rows;
         this.changeDetector.markForCheck();
     }
+
+    private _announceItems() {
```
gingi, line 132: It would be better to localize these strings. [Here's an example](https://github.com/Azure/BatchExplorer/blob/765a876fcda26eeab06390b8a0be67a7a82a6ef6/src/app/components/pool/action/edit/edit-app-package/edit-app-package-form.component.ts#L64) of how to localize parametrized strings.

gingi (commented):
Looks great! Great use of the `LiveAnnouncer` utility. Only suggestion is to localize the strings being announced.

gingi (approved):

wanghoppe:
@gingi  I made some changes to the unit test to pass the CI check, you might need to approve it again.

gingi (approved):

cRui861:
#2634

PR #2590, opened 2022-11-11
Lerna upgrade (fix for orphaned dev-server processes on Windows)

dpwatrous:
Lerna 4 leaked processes when using Ctrl-C to kill webpack-dev-server. Upgraded to Lerna 6 to resolve the issue.


zfengms (approved):

dpwatrous:
Update: Looks like this breaks Node 16 support so I'm going to have to go back to the drawing board (or we'll need to move to Webpack 5)

dpwatrous:
Removed the Node 18 support from this PR as it caused issues with older verisons of node.

PR #2602, opened 2022-12-09
Add mkdirp as an explicit dependency of desktop

dpwatrous:
This resolves one packaging issue where mkdirp wasn't included


dpwatrous:
Note that there is still a larger issue with packaging: @batch/* packages aren't bundled by electron-builder due to a limitation it has with symlinks.

gingi (approved):

PR #2605, opened 2022-12-15
feat: return React Component instead of render function in layout provider's getLayout 

wanghoppe:


dpwatrous:
We're renaming the `vnext` branch to `main` (and this will eventually replace `master` as the primary branch of the repo) so this PR will need to be rebased on `main`.

PR #2615, opened 2023-01-11
Only set build variables when running dist pipelines

dpwatrous:
This will make it so PR validation runs don't need permissions to change build tags, etc.


gingi (approved):

cRui861:
Fixes #2641

PR #2619, opened 2023-01-13
Fix multiple accessibility bugs.

wanghoppe:


gingi (changes requested):
Great work here. Requesting two general changes, which are fairly minor:

1. If an element only contains child elements with `aria-hidden="true"`, then the attribute should be set on that element.
2. Labels should be localized because they are announced by assistive technologies (as opposed to roles, which are in English).

gingi (approved):

gingi (approved):

cRui861:
Fixes #2634

PR #2624, opened 2023-01-21
Make form values read-only

dpwatrous:
Values should now be set through setValues() and updateValue() which will properly trigger change events


dpwatrous:
Previously, setting values like this wouldn't fire onChange:

```javascript
form.values.foo = "bar";
```

Now the above code won't compile, and values must be updated like so:

```javascript
// This will fire onChange
form.updateValue("foo", "bar");

// and so will this
form.getParam("foo").value = "bar";
```

gingi (approved):

PR #2629, opened 2023-01-25
Accessibility bug fixes

gingi:
Several other improvements:

* Systemic application of red asterisks for required fields (input directive only set it as a read-only property that did not honor validations or explicit `required` HTML attributes)
* Localization of several components
* Replacement of some custom colors with theme colors
* Corrected internal spelling of several classes, files, and directories

Most visible change is the removal of the light gray background of the square icon buttons.

Old:
MicrosoftTeams-image (2)

New:
image


dpwatrous (approved):

cRui861:
Fixes #2634

PR #2630, opened 2023-01-28
WIP: certificate list view with mobx

wanghoppe:


dpwatrous:
We're renaming the `vnext` branch to `main` (and this will eventually replace `master` as the primary branch of the repo) so this PR will need to be rebased on `main`.

PR #2633, opened 2023-02-01
Add localization build script, unit tests, and pipeline

skapur12:
Implement localization in Batch Explorer for all shared libraries packages and the desktop Electron application

- Add build script to iterate through all of the YAML files in a directory and generate an English JSON and RESJSON file
- Add unit tests to thoroughly test the build script
- Integrate OneLocBuild pipeline to generate files required for localization


packages/common/i18n/resources.resjson
```diff
@@ -0,0 +1,5 @@
+{
+  "lib.common.localizer.account": "Account",
+  "lib.common.localizer.resourceGroup": "Resource Groups",
```
gingi, line 3: Nit: Shouldn't this be a singular "Resource Group"?

gingi (commented):
Awesome work! Just one minor comment.

packages/common/i18n/resources.resjson
skapur12: Fixed this, thanks! (It doesn't show up here in the diff due to rebasing).

gingi (approved):

PR #2649, opened 2023-02-07
Bugs 462 + 290

robtaft-ms:
Hiding tooltip from the mini chart.
Changing form fields to using borders instead of grey backgrounds.


gingi:
Superseded by #2680

PR #2658, opened 2023-02-16
Update desktop localization script folder names for accuracy

skapur12:
**Changes to old localization system:**
- Change the old folder with translations originally named **i18n** in the desktop package to be **i18n-deprecated** (as it is no longer being updated).
- Modify the old desktop scripts that build the English string files to reference **i18n-deprecated** instead of **i18n**

**Changes to new localization system:**
- Change the new localization folder's name from **i18n-updated** to **i18n**
- Change new desktop localization script to point to **i18n** instead of **i18n-updated**


desktop/i18n-deprecated/resources.en.json
```diff
@@ -0,0 +1,384 @@
+{
+  "browse-layout.openedAdvancedFilter": "Opened advanced filter blade",
```
dpwatrous, line 2: Update .gitignore to ignore this file, as it's generated at compile-time.

dpwatrous: Actually all 3 of these lines in .gitignore need to be updated:

```
# Ignore the english compiled file
i18n/resources.en.json
i18n/xliff/
i18n/xliffs/
```

Note you'll need to do a `git rm` to remove this file from source control now that it's added.

dpwatrous (changes requested):

dpwatrous (approved):

PR #2665, opened 2023-02-27
Check that there are no resjson changes in Git

skapur12:
- Add a check to the pipeline to verify there weren't any resjson changes in Git that haven't been pushed
- (Basically, if a dev modifies any yml files, this check ensures that the developer has run `npm run build` to build translations and has pushed those changes with their code. If a dev makes changes to the yml files but doesn't run a build before merging their changes into main, the pipeline will fail.)


gingi (approved):
Looks great.

PR #2666, opened 2023-02-27
Form/action refactoring and other updates

dpwatrous:


gingi (approved):
Some opportunities for refactoring in the future, but this is great!

PR #2680, opened 2023-03-16
Accessibility fixes

gingi:
On behalf of Rob Taft.

### DIFF
--- a/src/@batch-flask/ui/form/form-field/form-field.scss
+++ b/src/@batch-flask/ui/form/form-field/form-field.scss
@@ -51,6 +51,7 @@ bl-form-field {
         width: 100%;
         flex: 1;
         min-width: 0;
+        border: 1px solid $input-border-color;
 
         &::placeholder {
             color: transparent;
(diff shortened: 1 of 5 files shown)


src/@batch-flask/ui/form/form-field/form-field.scss
```diff
@@ -51,6 +51,7 @@ bl-form-field {
         width: 100%;
         flex: 1;
         min-width: 0;
+        border: 1px #000000 solid;
```
cRui861, line 54: Could we switch this hex to use theme colors instead (like `$border-color`)? If we need to use black for this border then we could keep it since `$border-color` is currently set to a dark gray.

src/@batch-flask/ui/property-list/property-content/property-content.scss
```diff
@@ -2,14 +2,14 @@
...
     color: $primary-text;
-    background-color: $main-background;
     padding: 3px 8px;
     margin: 0;
     text-overflow: ellipsis;
     overflow: hidden;
     display: block;
     width: 100%;
     min-height: 24px;
+    border: 1px #000000 solid;
```
cRui861, line 12: Similarly with the hex here, could we switch it to use theme colors?

cRui861 (approved):

cRui861:
Fixes #2713

PR #2682, opened 2023-03-17
Update python dependencies

dpwatrous:


dpwatrous:
Note: This is marked as a draft because the Python server is already failing on master, and it doesn't make sense to update dependencies until that's fixed.

NickKouds (approved):

Issue #788, opened 2017-10-17
Pool start task failed quickfix not doing anything


timotheeguerin:
![image](https://user-images.githubusercontent.com/1031227/31690449-96603284-b347-11e7-9c01-ed82a3b0b978.png)
The fix start task button doesn't do anything

PR #2685, opened 2023-03-20
Accessibility fix for bug 788 ("delete metadata row" tooltip) -disregard

aishaBiber:
aibiber - Fix for #788 [Visual Requirement-Batch Explorer-Edit Metadata] Incorrect tooltip is define for delete metadata row icon under Edit metadata blade.


gingi:
Superseded by #2706

PR #2689, opened 2023-03-22
Keyboard accessibility bugs (463, 853)

mhdahman:
### Resolves:
 
- [Bug 853]: [Keyboard Navigation-Batch Explorer-SetUp]: Keyboard focus does not land on the first interactive control as soon as we invoke the 'Setup' control in Storage accounts pane.
- [Bug 463]: [Keyboard Navigation -Batch Explorer - Job preparation task] After activating ‘Job preparation task’ keyboard focus does not lands on first interactive control.

### Notable changes:

- Updated the `bl-complex-form` implementation so that it automatically focuses the first element in any `nested-form`s when they open. As a result, this fixes all instances of `bl-complex-form` that had similar auto focus issues, not just the "Job preparation task" form in the bug above. For example, a similar issue was found in the "Job manager task" nested form, which should be fixed now as well.
- Fixed runtime error causing keyboard tab navigation to stop working in the Storage accounts pane mentioned above


cRui861 (approved):

mhdahman:
@cRui861 I'm closing this PR to open a new one with a branch that I pushed directly to this repo instead of a fork. Sorry for the trouble 😄

PR #2693, opened 2023-03-23
Luminosity contrast accessibility bugs (568, 569)

mhdahman:
### Resolves:

- [Bug 568] [Visual Requirement-Batch Explorer-Add a batch account] Luminosity contrast ratio of the text "eastus.batch.azure.com" is less than 4.5:1 i.e 2.5:1.
- [Bug 569] [Visual Requirement-Batch Explorer-Node states] Luminosity contrast ratio of the text "Idle" is less than 4.5:1 i.e 2.5:1.

### Notable changes:

- Updated colors of monitor chart and other parts of the UI on the dashboard for both classic and dark themes to have luminosity contrast ratio >=4.5:1 — bug above was specific to classic theme
- Updated styles of bl-form-field to have luminosity contrast ratio >=4.5:1. This change affects all instances of `<bl-form-field>` that use `blFormFieldSuffix` or `blFormFieldPrefix`.
- Update tooltip configuration of the "Node states" chart to improve visibility. Please see my inline comment below for additional information.

--- 

Dashboard – classic theme:

Screen Shot 2023-03-23 at 1 32 56 PM

Dashboard – dark theme:

Screen Shot 2023-03-23 at 1 32 17 PM

Form fields with suffix – classic theme:

Screen Shot 2023-03-23 at 1 51 15 PM

Form fields with suffix – dark theme:

Screen Shot 2023-03-23 at 2 07 06 PM

### DIFF
--- a/data/themes/dark.json
+++ b/data/themes/dark.json
@@ -14,6 +14,18 @@
     "primary": "#cccccc",
     "secondary": "#a7a8a9"
   },
+  "monitorChart": {
+    "core-count": "#0A95FF",
+    "low-priority-core-count": "#AE8C00",
+    "task-start-event": "#AE8C00",
+    "task-complete-event": "#0A95FF",
+    "task-fail-event": "#F26363",
+    "starting-node-count": "#0A95FF",
+    "idle-node-count": "#C674D2",
+    "running-node-count": "#44A744",
+    "start-task-failed-node-count": "#F26363",
+    "rebooting-node-count": "#DB7843"
+  },
   "breadcrumb": {
     "text": "white",
     "background": "#5b5b5b",
--- a/src/app/components/account/monitoring/monitor-chart/monitor-chart.component.ts
+++ b/src/app/components/account/monitoring/monitor-chart/monitor-chart.component.ts
@@ -196,7 +196,8 @@ export class MonitorChartComponent implements OnChanges, OnDestroy {
             },
             tooltips: {
                 enabled: true,
-                mode: "index",
+                mode: "single",
+                position: "nearest",
                 callbacks: {
                     title: (tooltipItems, data) => {
                         return this._computeTooltipTitle(tooltipItems[0], data);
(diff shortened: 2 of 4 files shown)


src/app/components/account/monitoring/monitor-chart/monitor-chart.component.ts
```diff
@@ -196,7 +196,8 @@ export class MonitorChartComponent implements OnChanges, OnDestroy {
             },
             tooltips: {
                 enabled: true,
-                mode: "index",
+                mode: "single",
+                position: "nearest",
```
mhdahman, line 200: Update the tooltip configuration of monitor chart to improve the visibility of the tooltip. Please note that this changes the existing behavior of the tooltip to only show information about the hovered datapoint rather than displaying all datapoints at the hovered index which wasn't helpful because all datapoints could not be displayed within a limited space.

## Before

Screen Shot 2023-03-23 at 2 19 08 PM

Screen Shot 2023-03-23 at 2 19 03 PM

Screen Shot 2023-03-23 at 2 18 58 PM


## After

Screen Shot 2023-03-23 at 10 41 16 AM

Screen Shot 2023-03-23 at 10 40 50 AM

Screen Shot 2023-03-23 at 10 41 20 AM

data/themes/dark.json
```diff
@@ -14,6 +14,18 @@
     "primary": "#cccccc",
     "secondary": "#a7a8a9"
   },
+  "monitorChart": {
```
cRui861, line 17: Nit: Could we keep the lowercase letters since that was the format that already exists? Not that big of a deal but thought it would be better to remain consistent.

cRui861 (approved):

mhdahman:
@cRui861 Thanks for the review. I'm closing this PR to open a new one with a branch that I pushed directly to this repo instead of using a fork. Sorry for the trouble 😄

data/themes/dark.json
mhdahman: Good call! I'll address this in #2700

PR #2697, opened 2023-03-28
Theme colors will only show in dev mode

cRui861:
Add Developer menu item in Profile which would only appear while in dev mode. Fix some hardcoded color values to use our themes.


dpwatrous (approved):

cRui861:
Fixes #2714

PR #2699, opened 2023-03-31
Keyboard accessibility bugs (463, 853, 593)

mhdahman:
### Resolves:
 
- [Bug 853]: [Keyboard Navigation-Batch Explorer-SetUp]: Keyboard focus does not land on the first interactive control as soon as we invoke the 'Setup' control in Storage accounts pane.
- [Bug 463]: [Keyboard Navigation -Batch Explorer - Job preparation task] After activating ‘Job preparation task’ keyboard focus does not lands on first interactive control.
- [Bug 593] [Keyboard Navigation-Batch Explorer-Gallery-V-Ray] Visual focus indicator does not moves to the "run job with auto pool" button.

### Notable changes:

- Updated the `bl-complex-form` implementation so that it automatically focuses the first element in any `nested-form`s when they open. As a result, this fixes all instances of `bl-complex-form` that had similar auto focus issues, not just the "Job preparation task" form in the bug above. For example, a similar issue was found in the "Job manager task" nested form, which should be fixed now as well.
- Fixed runtime error causing keyboard tab navigation to stop working in the Storage accounts pane mentioned above

### DIFF
--- a/src/@batch-flask/ui/form/complex-form/complex-form.component.ts
+++ b/src/@batch-flask/ui/form/complex-form/complex-form.component.ts
@@ -1,5 +1,5 @@
 import {
-    AfterViewInit, ChangeDetectorRef, Component, ContentChildren, HostBinding, Input, OnChanges, QueryList, Type,
+    AfterViewInit, ChangeDetectorRef, Component, ContentChildren, ElementRef, HostBinding, Input, OnChanges, QueryList, Type, ViewChild,
 } from "@angular/core";
 import { FormControl } from "@angular/forms";
 import { AsyncTask, Dto, ServerError, autobind } from "@batch-flask/core";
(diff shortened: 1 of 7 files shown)


src/@batch-flask/ui/form/complex-form/complex-form.component.ts
```diff
@@ -268,4 +274,8 @@ export class ComplexFormComponent extends FormBase implements AfterViewInit, OnC
             multiUse: this.multiUse,
         };
     }
+
+    private focusFirstFocusableElement() {
+        this.formElement.nativeElement?.querySelector('[autofocus], button, input, textarea, select')?.focus()
+    }
```
mhdahman, line 280: The existing autofocus directive will not work as expected when used on an element nested inside of `complex-form` because those elements seem to be injected/rendered in the DOM before they're visible/shown. We'll have to explicitly focus those elements after the nested form page is opened.

cRui861 (approved):

mhdahman:
Pushed new changes to add a fix for `[Bug 593] [Keyboard Navigation-Batch Explorer-Gallery-V-Ray] Visual focus indicator does not moves to the "run job with auto pool" button`.

gingi (approved):

cRui861:
Fixes #2713

PR #2701, opened 2023-03-31
Add scripts to build localization on local machine

skapur12:
- Add npm `loc:restore`, `loc:build`, and `loc:clean` scripts to run `OneLocBuild `on local machine to install dependencies, build the translations, and correctly rename and move the translated files to their correct spot.
- Add logic to convert `RESJSON `into `JSON `files when localization is built and run locally (so we can use the `JSON `files instead for local development)
- Remove unnecessary `destJSON `parameter in build-translations script
- Update `package.json` files in package directories to reflect changes to build-translations script
- Add new README doc with instructions on how to set up localization on local machine


desktop/package.json
```diff
@@ -49,7 +49,7 @@
     "build:package": "npm run build:prod && npm run build-python && npm run package",
     "build:prod": "cross-env NODE_ENV=production npm run build",
     "build:test": "npm run build && npm run test",
-    "build-translations": "bux build-translations --src src --destJSON build/i18n-json --destRESJSON i18n",
+    "build-translations": "bux build-translations --src src --destRESJSON i18n",
```
dpwatrous, line 52: Can we make this just --dest if we only have one?

docs/build-localization-locally.md
```diff
@@ -0,0 +1,26 @@
...
+## Building Translations Files for Other Languages
+
+To build the localization translations for all languages besides English:
+
+* Follow the steps above first
+* Install the latest, recommended version of nuget.exe from <https://www.nuget.org/downloads> at C:\Users\{userName}, for instance.
+* Navigate to the root of the repository
+* Run `npm run loc:restore` to install all dependencies
+* Run `npm run loc:build` to build the translations and move them to their correct directories
+* If needed, run `npm run loc:clean` to clear out all previouslly installed dependencies
```
dpwatrous, line 21: I don't think that we need a clean to the NuGet dependencies, but it'd be great if when people ran `npm run clean` from the repo root, it would delete the directory which contains localization build output (`Localization/out` I think)

dpwatrous: Alternatively, `loc:clean` could just delete `Localization/out`

package.json
```diff
@@ -63,6 +63,9 @@
     "lint": "prettier -c . && lerna run --parallel workspace:lint --stream && npm run -s lint:markdown",
     "lint:fix": "prettier -w . && lerna run --parallel workspace:lint:fix --stream && npm run -s lint:markdown",
     "lint:markdown": "markdownlint-cli2 \"**/*.md\" \"#**/node_modules/**/*\" \"#SECURITY.md\"",
+    "loc:build": "cross-env-shell \"powershell -ExecutionPolicy Bypass -File ./Localize/build.ps1 && powershell -ExecutionPolicy Bypass -File ./Localize/copy-translations.ps1\"",
+    "loc:clean": "bux rmrf ./Localize/packages",
```
dpwatrous, line 67: (See other comment on cleaning localization build files)

dpwatrous (commented):
Looks really good overall - just a few comments

packages/common/package.json
dpwatrous, line 1: Let's add a "clean:resources" task in each of the packages so that when the "clean" task is run either from here or from the repo root, the "resources" directory is removed, along with all of the resjson/json output files.

package.json
dpwatrous, line 1: There are a lot of linting warnings from Prettier being generated by JSON files. Let's add the following to .prettierignore:

- Localize/out/
- Localize/packages/
- /**/resources/

dpwatrous (changes requested):

package.json
```diff
@@ -63,6 +63,9 @@
     "lint": "prettier -c . && lerna run --parallel workspace:lint --stream && npm run -s lint:markdown",
     "lint:fix": "prettier -w . && lerna run --parallel workspace:lint:fix --stream && npm run -s lint:markdown",
     "lint:markdown": "markdownlint-cli2 \"**/*.md\" \"#**/node_modules/**/*\" \"#SECURITY.md\"",
+    "loc:build": "cross-env-shell \"powershell -ExecutionPolicy Bypass -File ./Localize/build.ps1 && powershell -ExecutionPolicy Bypass -File ./Localize/copy-translations.ps1\"",
```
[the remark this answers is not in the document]
skapur12: Ah, nope, we aren't setting environment variables- I was just looking for a keyword that would let the script run on MacOS, Windows, and Linux and thought that cross-env-shell would work for that. How about  `pwsh -ExecutionPolicy Bypass -File ./Localize/build.ps1 && pwsh -ExecutionPolicy Bypass -File ./Localize/copy-translations.ps1`?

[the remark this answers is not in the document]
skapur12: I did this because the `pwsh `keyword doesn't work (it doesn't work on my Windows local box and also requires some overhead- I would need to download and install Powershell on Mac and Linux as well) and I wanted something that would allow the script to run cross-platform on Mac, Linux, and Windows. (Obviously `powershell `works only on Windows.) Is there another keyword I could use?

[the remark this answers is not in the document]
dpwatrous: This isn't going to work on Mac/Linux regardless, so IMO let's just remove cross-env-shell and call powershell directly.

dpwatrous (approved):

PR #2705, opened 2023-04-08
Fix for Dockerfile smell DL3048

grosa1:
Hi!
The Dockerfile placed at "scripts/docker/windows/Dockerfile" contains the best practice violation [DL3048](https://github.com/hadolint/hadolint/wiki/DL3048) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3048 occurs when the pattern used for the label keys does not match the format recommended by official guidelines.
In this pull request, we propose a fix for that smell generated by our fixing tool. We manually checked that the patch is correct before opening the pull request.
To fix this smell, specifically, the label keys are refactored to match the correct format.

This change is only aimed at fixing that specific smell. In the case the fix is not valid or useful, please briefly indicate the reason along with suggestions for possible improvements.

Thanks in advance.


gingi (approved):

gingi:
Hi @grosa1. Thanks for your contribution. Would you mind resolving the conflict in this PR? We have reorganized our codebase to accommodate new shared UI libraries. The original Batch Explorer project has been moved under `desktop/`, so you'll  need to move the Dockerfile to the appropriate path under that folder.

PR #2708, opened 2023-04-11
fix @azure/storage-blob usage

wanghoppe:


config/webpack.common.js
```diff
@@ -36,6 +36,13 @@ exports.defineEnv = function(env) {
             "RENDERER": JSON.stringify(true),
             "HOT": helpers.hasProcessFlag("hot"),
         },
+        // KLUDGE: This is a hack to make @azure/core-http thinks it's running in the browser,
+        // which uses process.versions.node to determine the environment.
+        // @azure/storage-blob uses isNode variable exported from @azure/core-http to determine
+        // how it should process data. Since we are using this SDK in the renderer process, it
+        // should get isNode = false.
+        // github issue: https://msazure.visualstudio.com/AzureBatch/_workitems/edit/17956715
```
dpwatrous, line 44: Can we link this issue instead? https://github.com/Azure/azure-sdk-for-js/issues/21110

dpwatrous (approved):
I think we should try to find a way to change the behavior of isNode without globally overwriting `process.versions.node`, but approving this PR so we can get a hotfix out in the insiders build in the meantime.

config/webpack.common.js
wanghoppe: Oh, sure. I mistake the ADO link as the GitHub issue line

patches/@azure+core-http+2.2.7.patch
```diff
@@ -0,0 +1,36 @@
...
++++ b/node_modules/@azure/core-http/dist/index.js
+@@ -317,10 +317,7 @@ const validUuidRegex = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F
+ /**
+  * A constant that indicates whether the environment is node.js or browser based.
+  */
+-const isNode = typeof process !== "undefined" &&
+-    !!process.version &&
+-    !!process.versions &&
+-    !!process.versions.node;
++const isNode = false;
```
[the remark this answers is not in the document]
dpwatrous: I think we should also add a test under `src/client` that checks that `isNode` ends up as `true`.

dpwatrous (approved):

dpwatrous:
Closing in favor of the PR to master

PR #2717, opened 2023-04-28
Ensures Batch Explorer can be packaged in the shared libraries

gingi:
electron-builder produces valid Batch Explorer release artifacts. The artifacts are inflated by the shared libraries, and they cannot be excluded in the config. The only way to exclude them is by moving them to `devDependencies` in `package.json` but that is the wrong approach.

Playground link in the profile menu is moved to the Developer submenu, which is only shown during development.


desktop/electron-builder.yml
```diff
@@ -17,7 +18,7 @@ extraResources:
       - "**/*"
   - "ThirdPartyNotices.txt"
   - "data"
-  - "i18n"
+  - "i18n-deprecated"
```
dpwatrous, line 21: Is this correct? I thought @skapur12 got it working with the new translation strings in the `i18n` directory.

dpwatrous (changes requested):

dpwatrous:
Looks good besides the test failure (which appears to just be an issue with a test that needs to be updated since the menus have changed)

desktop/electron-builder.yml
skapur12: The change to the desktop app is still on the dpwatrous/fake-localizer feature branch (not merged to main yet), I'll just include it in the next PR with all of the other changes!

dpwatrous (approved):

PR #2730, opened 2023-05-09
Integrate localized strings into web and desktop app

skapur12:
- Add localization support to web package
- Add a 'Merge Translations' node script to merge the shared libraries with web and desktop-specific translations and place them in web and desktop folders, respectively. It runs on every build
- Switch Electron app to use new localized strings
- Rename StandardLocalizer to BrowserLocalizer, and add error handling to BrowserLocalizer to fetch the appropriate file based on the locale in the user's browser
- Add global build-translations command to package.json
- Update README with instructions on setting up localization environment on dev box


util/bux/translation-functions.ts
```diff
@@ -13,10 +13,136 @@ type NestedStringMap<V> = StringMap<V> | StringMap<StringMap<V>>;
...
+    // Iterate through each resource directory
+    for (const dir of resourceDirs) {
+        // Iterate through each JSON file in the directory
+        for (const file of fs.readdirSync(dir)) {
+            if (file.startsWith("resources.") && file.endsWith(".json")) {
+                const langID = file.split(".")[1];
+
+                // If the language ID is not in the object, add it
+                if (
+                    !Object.prototype.hasOwnProperty.call(
```
dpwatrous, line 87: Could be `if (!mergedTranslations[langID]) {`

util/bux/translation-functions.ts
```diff
@@ -13,10 +13,136 @@ type NestedStringMap<V> = StringMap<V> | StringMap<StringMap<V>>;
...
+            path.join(rootDir, "packages/service/src/ui-service"),
+            path.join(rootDir, "packages/service/i18n"),
+            "lib.service"
+        )
+    );
+
+    await Promise.all(buildPackageEnglishPromises);
+
+    // Initialize an empty object to store the merged translations
+    const mergedTranslations = Object.create(null);
```
dpwatrous, line 76: Switch to `const mergedTranslations: Record<string, Record<string, string>> = {};` to get Typescript type checking

packages/react/src/ui-react/account/__tests__/create-account-action.spec.tsx
```diff
@@ -63,15 +63,16 @@ describe("Create account action", () => {
         await act(() => action.waitForInitialization());
         expect(action.isInitialized).toBe(true);
 
-        const accountNameInput: HTMLInputElement = await screen.findByLabelText(
-            "Account name",
+        const accountInput: HTMLElement[] = await screen.findAllByLabelText(
+            "Account",
```
dpwatrous, line 67: This should be "Account name"

dpwatrous: (That should let you revert the changes to this file)

dpwatrous (changes requested):

packages/react/src/ui-react/account/create-account-action.ts
```diff
@@ -52,18 +52,18 @@ export class CreateAccountAction extends AbstractAction<CreateAccountFormValues>
             values: initialValues,
         });
         form.param("subscriptionId", SubscriptionParameter, {
-            label: translate("subscription"),
+            label: translate("lib.common.localizer.subscription"),
```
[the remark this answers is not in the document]
skapur12: What if we had a `create-account-action.i18n.yml` file and added the strings to that (since the convention seems to give the yml file the same name as the Typescript file using it)? Not sure if the name might be too lengthy though

packages/react/src/ui-react/account/__tests__/create-account-action.spec.tsx
skapur12: This test fails with ‘Account name’ because the label of the element is actually ‘Account.’ Is the label actually supposed to be ‘Account name’?
create-account

.gitignore
```diff
@@ -13,4 +13,6 @@ Localize/out/
 /packages/playground/resources/i18n/*
 /packages/react/resources/i18n/*
 /packages/service/resources/i18n/*
-/util/bux/resources/i18n/json
+/util/bux/resources/i18n
```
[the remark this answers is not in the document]
dpwatrous: A couple options that would avoid needing to write a file that needs to be `.gitignore`'d specially for this test:

- Pass in an output path instead of `platform` to `createEnglishTranslations()` and write to a temp directory
or
- Refactor createEnglishTranslations() into 2 functions: one that returns a JSON string (or object) so that you can assert the translated data is correct, and one that writes that data to disk. The latter function could be tested separately by stubbing out the fs module to avoid a 'real' write, or just have it write to a temp file.

packages/react/src/ui-react/account/__tests__/create-account-action.spec.tsx
dpwatrous: Yup, the label should be "Account name" not "Account". This was the case before since the test was passing.

dpwatrous: Screenshot from the main branch:

Screenshot 2023-06-23 at 12 36 36 PM

desktop/src/app/app.component.ts
```diff
@@ -78,7 +78,7 @@ export class AppComponent implements OnInit, OnDestroy {
                 [DependencyName.Clock]: () => new StandardClock(),
                 // TODO: Create an adapter which hooks up to the desktop logger
                 [DependencyName.LoggerFactory]: () => createConsoleLogger,
-                [DependencyName.Localizer]: () => new StandardLocalizer(),
+                [DependencyName.Localizer]: () => new BrowserLocalizer(),
```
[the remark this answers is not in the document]
dpwatrous: Unresolving since this looks like it's still an issue

docs/build-localization-locally.md
```diff
@@ -1,26 +1,26 @@
...
-## Building the English File
+## Building the English File (Any Machine)
 
-To build the latest English translation resjson file from the YAML files:
+To build the latest English translation file from the YAML files:
 
-* Run `npm run build` to build translations for the entire repository
-* Run `npm run build-translations` in any package directory (desktop, packages/*) to build translations for a specific package
+* Run `npm run build-translations`
+* If you are removing lines from a YAML file, run `npm run loc:clean` first to clear out previously built translation files
```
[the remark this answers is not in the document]
dpwatrous: Unresolving

package.json
```diff
@@ -64,7 +65,7 @@
     "lint:fix": "prettier -w . && lerna run --parallel workspace:lint:fix --stream && npm run -s lint:markdown",
     "lint:markdown": "markdownlint-cli2 \"**/*.md\" \"#**/node_modules/**/*\" \"#SECURITY.md\"",
     "loc:build": "powershell -ExecutionPolicy Bypass -File ./Localize/build.ps1 && powershell -ExecutionPolicy Bypass -File ./Localize/copy-translations.ps1",
-    "loc:clean": "bux rmrf ./Localize/out",
+    "loc:clean": "bux rmrf ./Localize/out && bux rmrf ./packages/common/resources/i18n && bux rmrf ./packages/service/resources/i18n && bux rmrf ./packages/playground/resources/i18n && bux rmrf ./packages/react/resources/i18n && bux rmrf ./web/dev-server/resources/i18n && bux rmrf ./desktop/resources/i18n",
```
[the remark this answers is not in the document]
dpwatrous: Unresolving

packages/common/src/ui-common/__tests__/localization.spec.ts
```diff
@@ -7,7 +7,7 @@ describe("Localization utilities", () => {
     });
 
     test("Can return simple english strings", () => {
-        expect(translate("subscription")).toEqual("Subscription");
+        expect(translate("lib.common.localizer.account")).toEqual("Account");
```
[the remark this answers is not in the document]
dpwatrous: Unresolving

packages/common/src/ui-common/environment/__tests__/environment-util.spec.ts
```diff
@@ -150,7 +150,7 @@ describe("Environment tests", () => {
             super(config, {
                 clock: () => new FakeClock(),
                 loggerFactory: () => createMockLogger,
-                localizer: () => new StandardLocalizer(),
+                localizer: () => new BrowserLocalizer(),
```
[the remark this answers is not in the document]
dpwatrous: Unresolving

desktop/src/app/app.component.ts
```diff
@@ -68,17 +69,20 @@ export class AppComponent implements OnInit, OnDestroy {
         private ncjTemplateService: NcjTemplateService,
         private predefinedFormulaService: PredefinedFormulaService,
         private workspaceService: WorkspaceService,
+        private appLocaleService: AppLocaleService
     ) {
-        // Initialize shared component lib environment
-        initEnvironment(new DefaultBrowserEnvironment(
+        const localizer = new DesktopLocalizer(this.appLocaleService);
+        localizer.loadTranslations().then(() => {
```
dpwatrous, line 75: This looks like a race condition between isAppready and localization strings being loaded. Maybe use an Observable and put it into the combineLatest() list below?

dpwatrous: See haveWorkspacesLoaded in WorkSpaceService for an example

desktop/src/app/localizer/desktop-localizer.ts
```diff
@@ -0,0 +1,51 @@
...
+interface Translations {
+    [key: string]: string;
+}
+
+export class DesktopLocalizer implements Localizer {
+    private translations?: Translations;
+
+    constructor(private localeService: AppLocaleService) { }
+
+    async loadTranslations(): Promise<void> {
```
dpwatrous, line 16: Instead of implementing a loadTranslations() function here, can we just read the map on AppTranslationsLoaderService, which should already have all of the localized strings?

desktop/src/app/app.component.ts
dpwatrous: On second thought, if we can get rid of loadTranslations (see my other comment) then this doesn't need to involve async initialization at all.

gingi (approved):

desktop/src/client/main.ts
```diff
@@ -16,7 +16,12 @@ import "./init";
 // 2. Update electron user data folder
 import { parseArguments } from "./cli";
 const program = parseArguments(process.argv);
-import { app } from "electron";
+import { app, ipcMain } from "electron";
+
+// IPC Handler to get app path
+ipcMain.handle('get-app-path', async (event) => {
```
dpwatrous, line 22: Is this used anywhere?

packages/common/src/ui-common/localization/http-localizer.ts
```diff
@@ -0,0 +1,84 @@
...
+                    `Failed to load translations for '${languageToLoad}': ${error.message}`
+                );
+            } else {
+                throw error;
+            }
+        }
+    }
+
+    private async fetchTranslations(lang: string): Promise<Translations> {
+        const response = await fetch(`./resources/i18n/resources.${lang}.json`);
```
dpwatrous, line 62: The leading dot doesn't look right here since this is a relative URL and not a file path

packages/common/src/ui-common/localization/http-localizer.ts
```diff
@@ -0,0 +1,84 @@
+import { Localizer } from "./localizer";
+
+interface Translations {
+    [key: string]: string;
+}
+
+export class HttpLocalizer implements Localizer {
```
dpwatrous, line 7: Can we add a unit test for this file?

.gitignore
```diff
@@ -13,4 +13,5 @@ Localize/out/
 /packages/playground/resources/i18n/*
 /packages/react/resources/i18n/*
 /packages/service/resources/i18n/*
-/util/bux/resources/i18n/json
+web/dev-server/resources/i18n/*
```
dpwatrous, line 16: It looks like web/resources/i18n/* needs to be added .gitignore. The .json file is currently checked in to source control.

util/bux/package.json
```diff
@@ -19,7 +19,7 @@
   "scripts": {
     "build": "tsc -b ./config/tsconfig.build.json && node ./bin/bux chmodx ./bin/bux",
     "bux": "npm run build && node ./bin/bux",
-    "clean": "rimraf ./lib && rimraf ./build",
+    "clean": "rimraf ./lib && rimraf ./build && rimraf ./resources",
```
dpwatrous, line 22: It doesn't look like we need to clear out the resources directory here (at least I'm not seeing any resources directory being created when I build bux)

web/package.json
```diff
@@ -19,15 +19,17 @@
...
     "build:prod": "run-s clean compile bundle:prod",
     "build:test": "run-s build test",
+    "build-translations": "bux build-translations --src src/explorer-web --dest i18n --outputPath ./dev-server/resources/i18n",
     "bundle": "npm run bundle:dev",
     "bundle:analyze": "webpack --env.analyze",
     "bundle:dev": "webpack --env.dev",
     "bundle:prod": "webpack",
     "bux": "bux",
-    "compile": "tsc -b ./config/tsconfig.build.json",
+    "compile": "run-s build-translations && tsc -b ./config/tsconfig.build.json",
```
dpwatrous, line 28: `run-s` and `&&` are redundant here. I'd switch this to just `npm run build-translations && tsc -b ./config/tsconfig.build.json`

desktop/src/client/main.ts
skapur12: Nope! I removed it

.gitignore
skapur12: I don't see a web/resources directory, either generated during a build step or in source control: https://github.com/Azure/BatchExplorer/tree/dpwatrous/fake-localizer/web. 

(The web translations are stored in `web/dev-server/resources/i18n/*` since the server can only access resources inside the dev-server folder)

dpwatrous: Pretty sure this was my mistake due to switching branches and there's no issue here - sorry about that!

packages/common/src/ui-common/__tests__/http-localizer.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+import { CustomGlobal } from "./setup-tests";
+
+declare const global: CustomGlobal;
+
+describe("HttpLocalizer", () => {
+    let httpLocalizer: HttpLocalizer;
+    let fetchMock: jest.Mock;
+
+    beforeAll(() => {
+        const globalObject: CustomGlobal = global;
```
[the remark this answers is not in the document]
dpwatrous: Unfortunately it looks like `replaceProperty` requires a newer version of Jest than we're using, so another alternative would be to do something like this:

```
describe("HttpLocalizer", () => {
    const navigator: Mutable<typeof globalThis.navigator> = globalThis.navigator;
    const originalLanguage = globalThis.navigator.language;

    beforeEach(() => {
        navigator.language = "en-US";
    });

    afterAll(() => {
        navigator.language = originalLanguage;
    });

```

dpwatrous (changes requested):
The changes look good to me. I'm just going to block this until we've got a portal PR ready since it contains breaking changes.

dpwatrous (approved):

PR #2734, opened 2023-05-17
Reorganizes ESLint rules in BatchExplorer

gingi:
Overrides caused certain files from being ignored by ESLint. This PR places most of the rules into the desktop's rule set and ensures that it's used by all source code. Overrides define new rules, but still honor old rules. This PR also fixes a few dozen linting errors that were previously ignored (for example, missing semicolons from several unit test files).

Importantly, adds rules to disallow `fdescribe()` and `fit()`, so that we don't mistakenly merge focused tests.

### DIFF
--- a/desktop/.eslintrc.js
+++ b/desktop/.eslintrc.js
@@ -1,58 +1,21 @@
 module.exports = {
     ignorePatterns: [
         "node_modules/**/*",
-        "**/node_modules/**/*",
-        "lib/**/*",
-        "build/**/*",
-        "python/**/*",
-        "docs/**/*",
-        "data/**/*"
+        "/lib/**/*",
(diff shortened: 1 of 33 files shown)


desktop/.eslintrc.js
```diff
@@ -5,54 +5,18 @@ module.exports = {
         "lib/**/*",
         "build/**/*",
         "python/**/*",
```
dpwatrous, line 7: Should we add leading slashes to /build, /python, /lib, etc?

gingi: Done. It also makes me realize that `**/node_modules/**/*` is extraneous. Removed as well.

dpwatrous (approved):

PR #2738, opened 2023-06-01
Accessibility Fixes (332, 576, 592, 601, 598)

mhdahman:
This PR Fixes:

- [Bug 332] [Visual Requirement-Batch explorer-Pools] Luminosity contrast ratio of the focusable element is less than 3:1 i.e 1.2:1.
- [Bug 576] [Programmatic Access-Batch Explorer-Edit] On the Update the storage account blade, name and aria-expanded property are not defined for the expand and collapse button.
- [Bug 592] [Programmatic Access-Batch Explorer-Pools-Edit tags] On the pools blade, the name is not defined for the edit tags button.
- [Bug 601] [Programmatic Access-Batch Explorer-Pool-Nodes] The name is not defined for the toggle advanced filter button.
- [Bug 598] [Screen Reader-Batch Explorer-Jobs-Disable] ALT is not defined for the image available on the "Are you sure you want to disable job" dialog box.

I also [updated the colors](https://github.com/Azure/BatchExplorer/pull/2738/commits/846050792bdd9286e97e8bbda9e511850716859d) of elements in the sidebar selected item for better contrast ratios:

|Before|After|
|---|---|
|  Screenshot 2023-06-01 at 1 58 13 PM |  Screenshot 2023-06-01 at 1 53 45 PM  |
| Screenshot 2023-06-01 at 1 58 04 PM | Screenshot 2023-06-01 at 1 56 52 PM |

### DIFF
--- a/desktop/src/@batch-flask/ui/info-box/info-box.component.ts
+++ b/desktop/src/@batch-flask/ui/info-box/info-box.component.ts
@@ -2,11 +2,7 @@ import { Component, Input } from "@angular/core";
 
 @Component({
     selector: "bl-info-box",
-    template: `
-        <div class="infobox">
-            <div class="infobox-image"><i class="fa fa-info-circle fa-3" aria-hidden="true"></i></div>
-            <div class="infobox-text">{{message}}</div>
-        </div>`,
+    templateUrl: "info-box.html",
 })
 export class InfoBoxComponent {
     @Input()
--- a/desktop/src/@batch-flask/ui/list-and-show-layout/entity-details-list.html
+++ b/desktop/src/@batch-flask/ui/list-and-show-layout/entity-details-list.html
@@ -6,7 +6,8 @@
         <input blInput [formControl]="searchQuery" [placeholder]="filterPlaceholder + ' (Startwith)'"/>
         <bl-clickable *ngIf="enableAdvancedFilter"
             class="toggle-advanced-filter"
-            title="{{'entity-details-list.toggleAdvancedFilter' | i18n}}"
+            matTooltip="{{'entity-details-list.toggleAdvancedFilter' | i18n}}"
+            name="toggle"
             matTooltipPosition="above"
             [routerLink]="baseLink"
             [queryParams]="{filter: true}">
--- a/desktop/i18n/resources.resjson
+++ b/desktop/i18n/resources.resjson
@@ -149,6 +149,7 @@
   "gallery.recentTemplates": "Recent templates",
   "gallery.title": "Gallery applications",
   "gallery.viewReadme": "View readme",
+  "info-box.infoIconTitle": "Information",
   "job-progress-status.inaccurate": "Task count might not be accurate in real time",
   "job-progress-status.progressBar": "Job progress",
   "job-state.active.message": "Job is currently running. Task can be scheduled.",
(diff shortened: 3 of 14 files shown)


desktop/src/@batch-flask/ui/list-and-show-layout/entity-details-list.html
```diff
@@ -7,6 +7,8 @@
         <bl-clickable *ngIf="enableAdvancedFilter"
             class="toggle-advanced-filter"
             title="{{'entity-details-list.toggleAdvancedFilter' | i18n}}"
+            matTooltip="Toggle advanced filter"
```
gingi, line 10: I removed the `matTooltip` attribute, using the `title` attribute instead, for a very similar issue, AB#487. I think your change makes better sense, and it preserves the Material library's tooltip. Can you just remove the `title` attribute but also use the localized version of the string (the contents of the attribute value)?

desktop/src/@batch-flask/ui/info-box/info-box.component.ts
```diff
@@ -4,7 +4,7 @@ import { Component, Input } from "@angular/core";
     selector: "bl-info-box",
     template: `
         <div class="infobox">
-            <div class="infobox-image"><i class="fa fa-info-circle fa-3" aria-hidden="true"></i></div>
+            <div class="infobox-image"><i class="fa fa-info-circle fa-3" aria-hidden="true" title="Information"></i></div>
```
gingi, line 7: This should be localized. That's a bit tricky given that it's in a hardcoded template in the Component annotation. My suggestion is to extract the template into its own html file, then also create a localized file (`info-box.i18n.yml`). The `info-box.module.ts` would need to import `I18nUIModule` and the spec file would also need to import `I18nUITestingModule` for the tests to succeed. Let me know if you need help with this.

gingi (commented):
Just two requests. Ping me if you need a hand with those. Otherwise, looks great.

desktop/src/@batch-flask/ui/list-and-show-layout/entity-details-list.html
mhdahman: Will do! I'd also suggest using `matTooltip` for cases like this because it has a slight adverting over `title` as it makes tooltips visible on keyboard focus of interactive items, whereas `title` tooltips are only visible on mouse hover. `matTooltip` also adds a couple other `aria` attributes by default.

desktop/src/@batch-flask/ui/info-box/info-box.component.ts
mhdahman: Good call! I thought I was doing something wrong because the raw localization string (the yaml path) was showing up in the UI and I got stuck there trying to figure out what's going on, but then I decided to restart electron and it worked!

Looks like new localization files are not picked up until I restart? or am i still doing something wrong? 😅

Anyways, this should be localized now:

Screenshot 2023-06-02 at 3 30 25 PM

gingi: Yes, unfortunately, you have to restart Electron when a new file is introduced. But looks great. The only reason the build fails is that you need to run the translation build to update the strings file (`resources.resjson`). You can run `npm run build` from the root directory and commit the changes to the file.

gingi (approved):

PR #2744, opened 2023-06-08
Migrates AADGraph to MSGraph

gingi:
The only models using AADGraph are related to managing AAD applications (namely, `AADApplication` `ServicePrincipal`). Those now use MSGraph instead. AADGraph service modules are now marked as deprecated.


cRui861 (approved):

desktop/src/common/constants/constants.ts
```diff
@@ -109,7 +109,7 @@ export const ApiVersion = {
     },
     commerce: "2016-08-31-preview",
     authorization: "2017-05-01",
-    aadGraph: "1.6",
+    aadGraph: "1.6", // Deprecated
```
dpwatrous, line 112: Can we just remove all the AAD Graph code (and any related dependencies) instead of deprecating?

gingi: Sure! No actual reason to keep the code around.

dpwatrous (approved):

PR #2745, opened 2023-06-08
Fix AB#1003 Voiceover will now read aloud "+ create new file group"

cRui861:
- Also update resource group dropdown to use `bl-select` instead of `mat-autocomplete` in create new batch account


gingi (approved):
There are some failing unit tests, but other than making them pass, this looks great. Should make Batch Explorer forms more reliable and look better in general.

PR #2777, opened 2023-08-09
Add type safety to shared libraries localization

skapur12:
Add type safety to localization interface
- Add LocalizedStrings interface to common module that is augmented across all other packages. 
- Modified `build-translations` to generate Typescript interface for each package containing all strings for that package
- Switch the order of NPM commands so that `build-translations` now happens before Typescript compilation
- Add `clean:localization` command to clean out the generated Typescript files

(Note: I checked that these changes work with the portal by bux linking them and no changes need to be made to the portal- as we didn't change the localizer interface here and only the translate function, which isn't called yet in the portal ReactViews)


.gitignore
```diff
@@ -15,3 +15,9 @@ Localize/out/
 /packages/service/resources/i18n/*
 web/dev-server/resources/i18n/*
 desktop/resources/i18n/*
+/packages/common/src/ui-common/localization/generated/*
```
dpwatrous, line 18: Suggestion: Let's flip this around so `generated` is a top-level folder. This may have been a better place to put the generated .json files too now that I think about it...

packages/common/package.json
```diff
@@ -25,6 +25,7 @@
     "clean:build": "bux rmrf ./build",
     "clean:esm": "bux rmrf ./lib",
     "clean:cjs": "bux rmrf ./lib-cjs",
+    "clean:localization": "bux rmrf ./src/ui-common/localization/generated",
```
dpwatrous, line 28: If we do switch this so `generated` is at the top, this should probably be `clean:generated`

dpwatrous (changes requested):

packages/common/package.json
skapur12: I removed the `clean:localization` command altogether (as cleaning out the localization files via `workspace:clean` erases the translations after they are built, same issue that this PR fixed: https://github.com/Azure/BatchExplorer/pull/2775)

desktop/electron-builder.yml
```diff
@@ -20,7 +20,7 @@ extraResources:
       - "**/*"
   - "ThirdPartyNotices.txt"
   - "data"
-  - "resources"
+  - "src/generated/localization"
```
dpwatrous, line 23: This makes me think we should stick with `resources/*` or something similar for non-Typescript/Javascript files. I don't think we want our distributed packages to have a `src` directory which only contains `src/generated/localization` (which would include some Typescript files that may not compile). Let's follow up and talk about this, since I think we also need to differentiate between resources under source control (like *.css files, etc) and ones that are output artifacts of the build.

dpwatrous (changes requested):

desktop/electron-builder.yml
skapur12: Changed file structure so we're back to having the JSON/REJSON files in the `resources` directory and Typescript file in `generated/localization`. Also changed directory hierarchy to be `generated/localization` instead of `localization/generated` like it was previously

dpwatrous (approved):

PR #2778, opened 2023-08-09
Add notification API for shared libraries

cRui861:
Includes a notification demo in the playground for Portal and a mock notifier (for Batch Explorer).

**Screenshots from the Portal playground demo:**
- Notification popups.
image
- Notifications in the notification tray with the pending notification is in an updated status.
image
- Notifications in the notification tray with the pending notification in a completed status.
image

**Screenshots from web Playground demo**
- Using mock notification for alerts (will replace with a Batch Explorer Notifier)
image
image


packages/common/src/ui-common/environment/environment.ts
```diff
@@ -2,6 +2,7 @@ import { Clock } from "../datetime/clock";
 import { HttpClient } from "../http";
 import { Localizer } from "../localization";
 import type { Logger, LoggerFactory, LoggingContext } from "../logging";
+import { Notification as Notifier } from "../notification";
```
gingi, line 5: This is probably a pain, but maybe it's worth renaming "Notifcation" to "Notifier" systematically to distinguish a service that provides notifications from the notifications themselves and avoid confusion in the future.

gingi:
Rena, can you provide a screenshot or two of the demo?

gingi (changes requested):
The PR looks great, but I think renaming "Notification" to "Notifier" will avoid some confusion in the future.

packages/common/src/ui-common/environment/environment.ts
cRui861: Yeah good point, I'll update that too.

gingi (approved):

PR #2781, opened 2023-08-11
Shared library refactoring

dpwatrous:
This PR factors out common functionality which is not Batch-specific into two new packages: @azure/bonito-core and @azure/bonito-ui.


gingi (approved):

skapur12 (changes requested):
`npm run loc:build` is now failing due to invalid paths in `LocProject.json`. The `copy-translations` script also needs to be adjusted. Will add fixes for these

skapur12 (approved):

PR #2786, opened 2023-08-24
Minor fixes and descriptive comments

skapur12:
- Minor fixes/descriptive comments

Everything else (including the strings) has been moved to ClientResources in the portal PR: https://msazure.visualstudio.com/One/_git/Batch-UX-Portal/pullrequest/8654322


dpwatrous:
There are some dependency ordering issues here. The `bonito-*` packages don't depend on any of the Batch packages, and shouldn't contain any Batch-specific strings. Can we move all of this logic to the portal's build process? Doing it in this repo seems odd.

skapur12:
> There are some dependency ordering issues here. The `bonito-*` packages don't depend on any of the Batch packages, and shouldn't contain any Batch-specific strings. Can we move all of this logic to the portal's build process? Doing it in this repo seems odd.

The logic that compiles the shared libraries package translations together has now been moved out of this PR and into the portal PR: https://msazure.visualstudio.com/One/_git/Batch-UX-Portal/pullrequest/8654322
This PR just contains the localizable strings needed for that PR

dpwatrous (approved):

PR #2787, opened 2023-08-24
Require Node.js >= 18

dpwatrous:
Note this will prevent builds from working when using Node < 18. This is because the `NODE_OPTIONS=--openssl-legacy-provider` workaround to let Webpack 4 work for Node 18 is incompatible with older versions.


dpwatrous:
This is failing on the CI builds but I can't seem to reproduce it locally. Still looking into it.

dpwatrous:
Abandoning this for now. Will recreate when I figure out the CI issues.

PR #2798, opened 2023-09-08
Add test-axe script for desktop

cRui861:


desktop/package.json
```diff
@@ -25,6 +25,7 @@
     "clean": "rimraf build/* dll/* python/dist/* python/build/* src/generated/*",
     "karma": "cross-env NODE_OPTIONS=--openssl-legacy-provider node --max_old_space_size=4096 node_modules/karma/bin/karma",
     "test": "npm run test-client:coverage && npm run test-app",
+    "test-axe": "cross-env BE_ENABLE_A11Y_TESTING=true npm run test-app",
```
dpwatrous, line 28: Can we make this `test-app:a11y` to match the other scripts? Also, let's add `test:all` and have `workspace:test:all` call it. We need one command for CI builds that runs all of the unit tests including coverage and a11y tests.

dpwatrous (changes requested):

desktop/package.json
cRui861: Updated

dpwatrous (approved):

dpwatrous:
Looks good!

PR #2803, opened 2023-09-16
Dataplane update

cRui861:
Update properties for models. Also added type definitions for JSON for issues with only an `object` type


gingi (approved):

cRui861:
Fixes #2834

PR #2810, opened 2023-10-16
Enable node communication mode editing

dpwatrous:
This was only enabled in dev mode before. Now that we've updated to the latest API version, we can enable it in production.


dpwatrous:
Screenshots of the pool configuration view now that it can display real data:

Minimized with different current/target values:
Screenshot 2023-10-16 at 11 35 12 AM

Minimized with same or default values:
Screenshot 2023-10-16 at 11 35 22 AM

Expanded:
Screenshot 2023-10-16 at 11 35 31 AM

cRui861 (approved):

PR #2815, opened 2023-10-18
fix certificate display list issue

wanghoppe:


packages/react/src/components/certificate/certificate-list.tsx
```diff
@@ -28,7 +28,8 @@ export const CertificateList = observer((props: CertificateListProps) => {
                 { prop: "state", label: "State" },
                 { prop: "stateTransitionTime", label: "Last Updated" },
             ]}
-            items={view.items}
+            items={[...view.items]}
```
dpwatrous, line 31: As we talked about earlier, let's get rid of the shallow clone here and just change how updating items in the list view works.

dpwatrous (changes requested):

dpwatrous (approved):

PR #2822, opened 2023-10-24
WIP: add validation data to paramter and controls

wanghoppe:


packages/bonito-core/src/form/validation-status.ts
```diff
@@ -1,13 +1,15 @@
...
     forced?: boolean = false;
 
-    constructor(level: "ok" | "warn" | "error" | "canceled", message?: string) {
-        this.level = level;
-        this.message = message;
-    }
+    constructor(
+        public level: "ok" | "warn" | "error" | "canceled",
+        public message?: string,
+        public data?: VD
```
dpwatrous, line 13: What do you think about VD meaning ValidationDetails instead of ValidationData? It's not much more specific, but somewhat matches Batch's error pattern of message + details.

packages/bonito-core/src/form/form.ts
```diff
@@ -75,13 +76,13 @@ export interface Form<V extends FormValues> {
 
     param<
         K extends ParameterName<V>,
-        D extends ParameterDependencies<V> = ParameterDependencies<V>,
-        T extends Parameter<V, K, D> = Parameter<V, K, D>
+        D extends ParameterDependencies<V> = NoDependencies,
+        VD = unknown
```
dpwatrous, line 80: I think this should default to `never` instead of `unknown`

packages/bonito-ui/src/components/form/string-list.tsx
```diff
@@ -0,0 +1,170 @@
...
+                onClick={() => {
+                    onDelete(index);
+                }}
+                disabled={disableDelete}
+            />
+        </Stack>
+    );
+}
+
+function useStringListItemStyles(props: StringListItemProps) {
```
dpwatrous, line 135: Should we switch this to use mergeStyles?

dpwatrous (changes requested):

packages/bonito-core/src/form/validation-status.ts
wanghoppe: Like it, change to ValidationDetials

packages/bonito-ui/src/components/form/string-list.tsx
wanghoppe: `mergeStyles` create class names, which needs to add more wrapper `div` around FluentUI components and then apply to it. I found it more convenient to just apply styles to the FluentUI components.

wanghoppe: Btw, if we want to modify the FluentUI components' internal element's style, we need to pass style object to it

PR #2825, opened 2023-10-27
add validation details and string-list control

wanghoppe:


packages/bonito-ui/src/components/form/string-list.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+function useStringListItemStyles(props: StringListItemProps) {
+    const theme = useAppTheme();
+    const { disableDelete } = props;
+
+    const itemPadding = {
+        padding: "11px 8px 11px 12px",
+    };
+
+    return {
```
dpwatrous, line 147: I see your point about overriding styles, though in general it makes me a bit nervous to do that as specific inner styles tend to be implementation details which change between releases.

Can we at least memoize this function? These nested style objects are being created once for every textbox in the list per keystroke right now regardless of whether anything changed.

dpwatrous (changes requested):

packages/bonito-ui/src/components/form/string-list.tsx
wanghoppe: Actually using `styles` props to customize component style is what they recommended. For example for `TextField`, there is a [interface](https://developer.microsoft.com/en-us/fluentui#/controls/web/textfield#ITextFieldStyles) to show what part of a `TextField` you can customize. I think it will be stable in Major versions. Fluent V9 is another story, they don't recommend to use `mergeStyles` any more, see: https://react.fluentui.dev/?path=/docs/concepts-migration-from-v8-components-theme-migration--page#mergestyles---mergeclasses

wanghoppe: Agreed that we should memorize this function

packages/bonito-ui/src/components/form/form-control.ts
```diff
@@ -63,5 +63,5 @@ export interface FormControlProps<
     /**
      * Callback for when the value of the control changes
      */
-    onChange?: (event: React.FormEvent, value: V[K]) => void;
+    onChange?: (event: React.FormEvent | null, value: V[K]) => void;
```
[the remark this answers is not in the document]
wanghoppe: Haven't tried in `StringList` in Portal yet, will try now. Found an example in their FormBuilder [docs](https://eng.ms/docs/products/azure-portal-framework-ibizafx/reactviews/mainconcepts/formbuilder/apireference#customization-options-2):
image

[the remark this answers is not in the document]
wanghoppe: Tested in Portal, `StringList` control works fine with Portal's form builder API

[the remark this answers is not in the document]
dpwatrous: Gotcha, thanks!

packages/bonito-ui/src/components/form/string-list.tsx
dpwatrous: Jeez, Fluent UI 8 -> 9 is going to be a ton of fun (whenever the portal supports 9). I'm fine with keeping it and just adding some memoization.

dpwatrous (approved):

PR #2847, opened 2023-11-17
AB#25538845 Add Chromium to 3rd party notice file, update dependency …

ReneOv-MSFT:
- Added Chromium to additional 3rd party dependencies for its license to be generated and added to notice file with script.
- Updated [release wiki page](https://msazure.visualstudio.com/AzureBatch/_wiki/wikis/AzureBatch.wiki/10538/Deploy-a-new-version-of-Batch-Explorer) with new step to manually review additional dependencies' versions.

### DIFF
--- a/desktop/scripts/lca/generate-third-party.ts
+++ b/desktop/scripts/lca/generate-third-party.ts
@@ -61,7 +61,7 @@ const additionalDependencies: Dependency[] = [
     },
     {
         name: "node.js",
-        version: "7.4.0",
+        version: "20.9.0",
         url: "https://nodejs.org/en/",
         repoUrl: "https://github.com/nodejs/node",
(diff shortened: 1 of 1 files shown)


desktop/scripts/lca/generate-third-party.ts
```diff
@@ -61,7 +61,7 @@ const additionalDependencies: Dependency[] = [
     },
     {
         name: "node.js",
-        version: "7.4.0",
+        version: "20.9.0",
```
dpwatrous, line 64: Let's make this match the version that Electron ships with.

dpwatrous (changes requested):

dpwatrous (approved):

cRui861:
Fixes #2875

PR #2884, opened 2024-01-16
feat: add vm extension related components & enable transpiling ESM js file for jest

wanghoppe:
image

image


packages/react/src/vm-extension/mock-vme-ext-items.ts
gingi, line 1: Should this be named `mock-vm-ext-items.ts`?

gingi (dismissed):
Looks good. One small question about a file name; if by intention then ignore.

packages/react/src/vm-extension/mock-vme-ext-items.ts
wanghoppe: Yeah, thanks for catching that, will fix.

packages/bonito-ui/src/components/property/property-field.tsx
```diff
@@ -6,6 +6,8 @@ import { copyToClipboard } from "@azure/bonito-core";
 export interface PropertyFieldProps<T> {
     label?: string;
     value?: T;
+    hideCopyButton?: boolean;
+    labalStyle?: React.CSSProperties;
```
dpwatrous, line 10: labal -> label

packages/bonito-ui/src/environment/mock-browser-environment.ts
```diff
@@ -38,6 +38,24 @@ export class MockBrowserEnvironment
...
+        // some JSDOM APIs are not implemented, so we need to mock them
+        // to make importing monaco-editor works
+        // https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom
+        Object.defineProperty(window, "matchMedia", {
+            writable: true,
+            value: jest.fn().mockImplementation((query) => ({
+                matches: false,
+                media: query,
+                onchange: null,
+                addListener: jest.fn(), // deprecated
```
dpwatrous, line 51: Let's move this into Jest-specific code. MockBrowserEnvironment can also be used for the playground, and doesn't require Jest. It also runs in a real browser, so doesn't need these stubs.

packages/playground/src/demo/display/vm-extension/vm-extension-list-demo.tsx
```diff
@@ -0,0 +1,32 @@
...
+    // const [panelOpen, setPanelOpen] = React.useState<boolean>(false);
+    // const [selectedExtension, setSelectedExtension] =
+    //     React.useState<VmExtItem>();
+
+    return (
+        <DemoPane title="VM Extension List">
+            <VmExtensionList
+                extensions={allExtItems}
+                loading={false}
+                // onItemClick={(item) => {
```
dpwatrous, line 20: Can this be removed?

packages/react/src/vm-extension/__tests__/node-vm-ext-list.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+    it("should render vm extension lists", async () => {
+        const { getByText } = render(
+            <NodeVMExtList
+                accountEndpoint="dummy_endpoint"
+                poolId={"dummy_pool"}
+                nodeId="tvmps_id1"
+            />
+        );
+        getByText(translate("lib.react.vmExtension.name"));
```
dpwatrous, line 20: Can we add an accessibility test in here somewhere? Also in the pool extension list test.

packages/react/src/vm-extension/node-vm-ext-list.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const nodeService: NodeService = React.useMemo(() => {
+        return inject(BatchDependencyName.NodeService);
+    }, []);
+
+    React.useEffect(() => {
+        setLoading(true);
+        nodeService
+            .listBatchNodeExtensions(accountEndpoint, poolId, nodeId)
+            .then((resList) => {
+                const exts = [];
```
dpwatrous, line 29: Need to check if the component is still mounted by the time this runs. To track this, you can return a callback which sets a local variable. See action-form.tsx for an example.

This is a pattern that's really hard to get right consistently. Probably a good case for a custom hook.

dpwatrous (changes requested):

packages/react/src/vm-extension/node-vm-ext-list.tsx
dpwatrous: To expand a bit on the issue, imagine the following hook:

```
useEffect(() => {
    loadSomeData()
        .then((data) => {
            setSomeState(data);
        });
}, [dependency1, dependency2])
```

Now imagine the hook is run multiple times in quick succession due to component rerendering. Every `loadSomeData()` promise might still be unresolved, and there's no guaranteed order in which they'll resolve. The callbacks will race, and the final data passed to `setSomeState()` will vary based on the order in which the callbacks fire.

Adding a check to see if the component is still mounted before setting state will resolve this:

```
useEffect(() => {
    let isMounted = true;
    loadSomeData()
        .then((data) => {
            if (isMounted) {
                setSomeState(data);
            }
        });
    return () => { isMounted = false; };
}, [dependency1, dependency2])
```

See also: https://react.dev/learn/synchronizing-with-effects#fetching-data

packages/bonito-core/config/tsconfig.test.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.common.json",
+  "compilerOptions": {
+    "tsBuildInfoFile": "../build/cache/ts.build.buildinfo",
```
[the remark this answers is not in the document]
wanghoppe: renamed

util/common-config/jest-common.js
```diff
@@ -55,6 +59,12 @@ module.exports = {
                 ],
             ],
             globals: {
+                "ts-jest": {
+                    // js-with-ts preset require "allowJs": true in tsconfig.json
+                    // so making a "tsconfig.test.json" to override it
+                    tsconfig: "config/tsconfig.test.json",
+                    isolatedModules: true,
```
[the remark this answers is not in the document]
wanghoppe: Tested it only works in `globals` with `js-with-ts` preset

util/common-config/jest-common.js
```diff
@@ -105,8 +119,15 @@ module.exports = {
         baseConfig.moduleNameMapper["@batch/ui-service/lib/(.*)$"] =
             "@batch/ui-service/lib-cjs/$1";
 
+        // css files
+        baseConfig.moduleNameMapper["^.+\\.css$"] = path.join(
+            __dirname,
+            "../mock-style.js"
```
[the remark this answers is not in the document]
wanghoppe: Added comments

packages/bonito-ui/src/components/panel/panel.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export interface PanelProps {
+    headerText?: string;
+    isOpen?: boolean;
+    onDismiss?: () => void;
+    isBlocking?: boolean;
+    footerButtons?: ButtonProps[];
+}
+
+export const Panel: React.FC<PropsWithChildren<PanelProps>> = (props) => {
```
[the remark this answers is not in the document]
wanghoppe: Added comments

packages/bonito-ui/src/components/property/__tests__/property-group.spec.tsx
```diff
@@ -56,4 +57,28 @@ describe("PropertyGroup component", () => {
         expect(heading.getAttribute("aria-expanded")).toBe("false");
         expect(section.style.display).toBe("none");
     });
+
+    test('Should not be able to expand/collapse if "enableCollapse" is false', async () => {
+        await render(
+            <PropertyGroup title="Car" enableCollapse={false}>
```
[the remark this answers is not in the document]
wanghoppe: Done

packages/bonito-ui/src/environment/mock-browser-environment.ts
wanghoppe: moved jsdom mocks to setup-tests

packages/playground/src/demo/display/vm-extension/vm-extension-list-demo.tsx
wanghoppe: Resolved

packages/react/src/vm-extension/__tests__/node-vm-ext-list.spec.tsx
wanghoppe: added for node and pool VM extension list

packages/bonito-ui/src/components/data-grid/data-grid.tsx
```diff
@@ -102,7 +102,14 @@ export interface DataGridColumn {
     /**
      * Custom renderer for cell content, instead of the default text rendering.
      */
-    onRender?: (item?: unknown, index?: number, column?: IColumn) => unknown;
+    onRender?: (
+        // can't use unknown here because it's a function's parameter, typescript
+        // check it contravariantly
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        item: any,
```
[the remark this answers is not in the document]
wanghoppe: add generic type for data-grid

packages/bonito-core/src/__tests__/setup-tests.ts
```diff
@@ -1,15 +1,4 @@
 import { destroyEnvironment } from "../environment";
```
dpwatrous, line 1: I'm all for factoring out our JSDOM stubs/mocks, but at least some of that needs to be in bonito-core since HttpClient needs both fetch() and Blob.

packages/bonito-ui/src/environment/mock-browser-environment.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
```
dpwatrous, line 1: Why is this needed?

util/mock-style.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
dpwatrous, line 1: Can we move this into common-config? `util` wasn't meant to contain source code, but rather be a parent folder for utility packages.

packages/react/src/vm-extension/mock-vm-ext-items.ts
```diff
@@ -0,0 +1,92 @@
+import { VmExtItem } from "./vm-extension-list";
+
+export const succeededExtItem: VmExtItem = {
```
dpwatrous, line 3: Can we move this file to the __tests__ folder if this isn't referenced outside of the local tests?

dpwatrous (changes requested):

packages/bonito-core/src/__tests__/setup-tests.ts
wanghoppe: Yeah, that's the problem, do you think it's OK to move `jsdom-mocks.ts` into bonito-core or just the blob and fetch?

packages/bonito-ui/src/components/data-grid/data-grid.tsx
```diff
@@ -102,7 +102,14 @@ export interface DataGridColumn {
     /**
      * Custom renderer for cell content, instead of the default text rendering.
      */
-    onRender?: (item?: unknown, index?: number, column?: IColumn) => unknown;
+    onRender?: (
+        // can't use unknown here because it's a function's parameter, typescript
```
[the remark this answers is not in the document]
wanghoppe: will remove

packages/bonito-ui/src/environment/mock-browser-environment.ts
wanghoppe: will remove

packages/react/src/vm-extension/mock-vm-ext-items.ts
wanghoppe: This is used in playground as well to display VM extension list demo page, so distributing it as well.

packages/bonito-core/src/__tests__/setup-tests.ts
dpwatrous: I'm a little torn, but if we want to keep saying that bonito-core can work in Node.js we should probably split them. Node does have Blob and fetch, but won't likely ever implement `matchMedia` or `HTMLCanvasElement`.

packages/react/src/vm-extension/mock-vm-ext-items.ts
dpwatrous: Ah, I missed that. Can we move this to a folder which makes it clear this is for tests? Could be either the service package's test-util folder, or one in the react package. Just want to be careful mixing test data and code with production code (even though I know there's an overlap when it comes to the playground).

wanghoppe: moved to @batch/ui-react package's test-util

packages/react/src/vm-extension/__tests__/vm-extension-list.spec.tsx
```diff
@@ -0,0 +1,90 @@
+// FILEPATH: /c:/Users/hoppe/work/sandbox/BatchExplorer/packages/react/src/vm-extension/__tests__/vm-extension-list.spec.tsx
```
dpwatrous, line 1: Remove this line

packages/react/src/test-util/mock-vm-ext-items.ts
```diff
@@ -0,0 +1,92 @@
+import { VmExtItem } from "../vm-extension/vm-extension-list";
+
+export const succeededExtItem: VmExtItem = {
```
dpwatrous, line 3: Seems like we should have some extension data in BatchFakeSet

packages/react/src/vm-extension/__tests__/vm-extension-details-panel.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+        const { getByText } = render(
+            <VmExtensionDetailsPanel
+                isOpen={false}
+                vme={succeededExtItem}
+                onDismiss={onDismiss}
+            />
+        );
+        expect(() =>
+            getByText(translate("lib.react.vmExtension.extensionProperties"))
+        ).toThrow();
```
dpwatrous, line 36: Can we assert positively that there's an empty or hidden element or something like that instead of just checking that the text doesn't exist?

wanghoppe: This one is hard. When the panel opens, it actually adds a <div> under <body> and removes it when closes.

dpwatrous: Happy to follow up on this, but this isn't a total blocker. I'll just resolve and approve since this PR has been out for so long.

dpwatrous (approved):

util/common-config/jest-common.js
```diff
@@ -55,6 +59,14 @@ module.exports = {
                 ],
             ],
             globals: {
+                "ts-jest": {
+                    // js-with-ts preset require "allowJs": true in tsconfig.json
+                    // so making a "tsconfig.test.json" to override it
+                    tsconfig: "config/tsconfig.test.json",
```
dpwatrous, line 65: Can we either change the `createConfig()` function to not take in a tsconfig, or pass `config/tsconfig.test.json` in for each package?

dpwatrous (approved):

util/common-config/jest-common.js
wanghoppe: We are passing a `tsconfig` object read from `tsconfig.json` for each package, and the only usage of this object is to pass `compilerOptions.paths` into jest's `moduleMapper`. But I didn't find any `tsconfig.json` we read from has the `compilerOptions.paths` field, do you think it's OK to remove the `tsconfig` object argument for `createConfig()`?

dpwatrous: It looks like `ts-jest`'s tsconfig property supports an inline object, so I guess I'd lean toward using the passed-in object. Then we'd just have to update each jest.config.js from `require("./tsconfig.json")` to `require("./config/tsconfig.test.json")`. That also keeps the `config/ts*.json` directory structure a convention rather than something jest-common.js needs to work.

wanghoppe: Removed `tsconfig` parameter for `createConfig()`

dpwatrous (dismissed):

dpwatrous (approved):

PR #2891, opened 2024-03-22
Rendering UI removal

cRui861:
The pool create form no longer has the option to select a graphics or rendering OS and the application licenses have been removed.

Before:
image

After:
image

The gallery menu item remains but only holds a deprecation warning regarding the retirement of the graphics and rendering pools.

Before:
image

After:
image


desktop/src/app/components/gallery/home/gallery.html
```diff
@@ -2,36 +2,13 @@
...
-        <bl-local-template-pick-button></bl-local-template-pick-button>
     </div>
     <div class="gallery-content">
-        <div class="gallery-applications">
-            <div class="filter">
-                <input autofocus blInput [formControl]="quicksearch" placeholder="Search">
+        <bl-banner class="gallery-deprecation-warning" type="warning">
+            <div message>
+                Graphics and rendering pools have been retired on 29 February 2024.
+                <a class="gallery-warning-alternative-link" (click)="openLink('https://azure.microsoft.com/en-us/updates/azure-batch-rendering-vm-images-licensing-will-be-retired-on-29-february-2024/')" href="javascript:void(0)">Click for more info</a>
```
dpwatrous, line 10: Remove "en-us" from the URL so it can be localized:

https://azure.microsoft.com/updates/azure-batch-rendering-vm-images-licensing-will-be-retired-on-29-february-2024/

dpwatrous (changes requested):

desktop/src/app/components/gallery/home/gallery.html
cRui861: Updated

desktop/src/app/components/gallery/home/gallery.html
```diff
@@ -2,36 +2,13 @@
...
-        <div class="flex-separator"></div>
-        <bl-local-template-pick-button></bl-local-template-pick-button>
     </div>
     <div class="gallery-content">
-        <div class="gallery-applications">
-            <div class="filter">
-                <input autofocus blInput [formControl]="quicksearch" placeholder="Search">
+        <bl-banner class="gallery-deprecation-warning" type="warning">
+            <div message>
+                Gallery applications are no longer supported as of February 29, 2024. See the Azure Batch rendering VM images & licensing retirement announcement for details.
```
dpwatrous, line 9: Can we extract this out to a localized string? If we switch the link text to "Learn more" we can use `common.learnMore` for that, which is already localized.

dpwatrous: We should also parameterize the date here. See `pool-create-basic-dialog.deprecationWarning.imageNearingEndOfLife` for an example.

cRui861: Updated

desktop/src/app/components/gallery/home/gallery.component.ts
```diff
@@ -32,6 +32,8 @@ export class GalleryComponent implements OnInit, OnDestroy {
     private _baseUrl: string;
     private _destroy = new Subject();
 
+    public deprecateRenderingDate = new Date(2024, 1, 29).toDateString()
```
dpwatrous, line 35: Can we use a localized date string here? If we don't have any good utility functions for this, we may want to add something to bonito-core

dpwatrous (approved):

PR #2893, opened 2024-04-04
Update prettier to 3.x

dpwatrous:
Also contains a random package-lock.json update due to merging from stable to main.


gingi (approved):

dpwatrous:
Updating was causing issues with markdownlint because of prettier's change to format *.jsonc files with trailing commas. Added overrides to prevent this. See: https://github.com/prettier/prettier/issues/15956

gingi (approved):

PR #2908, opened 2024-05-30
Add bcoroneliit to CODEOWNERS

ReneOv-MSFT:


gingi (approved):

skapur12 (changes requested):
BTW, just thought I'd mention that Dave mentioned to me that we weren't going to add interns to codeowners because you only need to be in codeowners to approve PRs and we can approve their PRs- but let me know if that's not the case

PR #2912, opened 2024-06-07
Update Batch Dataplane RLC with latest TypeSpec

wanghoppe:


dpwatrous:
Looks good to me, let's just make sure to remove the Typescript/Angular kludge when we're able to update Angular.

PR #2916, opened 2024-06-21
Build scripts call python3/pip3 instead of python/pip

dpwatrous:
This fixes an issue with the Mariner build container which doesn't provide python/pip on the path.


dpwatrous:
Closing this, since I resolved it in the official build pipeline instead of here.

PR #2920, opened 2024-07-02
Add AccountService and CacheManager

wanghoppe:


packages/bonito-core/src/cache/cache-utils.ts
```diff
@@ -0,0 +1,11 @@
+import { getEnvironment } from "../environment";
+import { CacheManager } from "./cache-manager";
+
+/**
+ * Gets the notifier for the current environment
```
dpwatrous, line 5: Should be 'Gets the cache manager'

packages/bonito-core/src/environment/environment.ts
```diff
@@ -56,6 +57,11 @@ export interface Environment<C extends EnvironmentConfig> {
      */
     getNotifier(): Notifier;
 
+    /**
+     * Gets the notifier for the current environment
```
dpwatrous, line 61: Notifier -> cache manager

packages/react/src/vm-extension/node-vm-ext-list.tsx
```diff
@@ -21,38 +21,58 @@ export const NodeVMExtList = (props: NodeVmExtensionListProps) => {
         return inject(BatchDependencyName.NodeService);
     }, []);
 
+    const accountService: AccountService = React.useMemo(() => {
```
dpwatrous, line 24: This is totally fine, but FYI there's not much reason to use `useMemo()` here as `inject()` is a very cheap function and should return a stable value.

dpwatrous: (Feel free to resolve this with no change - just wanted to point this out)

packages/react/src/vm-extension/node-vm-ext-list.tsx
```diff
@@ -21,38 +21,58 @@ export const NodeVMExtList = (props: NodeVmExtensionListProps) => {
...
+        return inject(BatchDependencyName.AccountService);
+    }, []);
+
+    const fetchExtensions = React.useCallback(async () => {
+        const accountData = await accountService.get(accountResourceId);
```
dpwatrous, line 29: Should we just add a `getRequired()` function to AccountService which throws if no account is found?

packages/service/src/pool/live-pool-service.ts
```diff
@@ -20,10 +19,7 @@ const SINGLE_POOL_PATH =
 const POOLS_PATH =
     "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Batch/batchAccounts/{accountName}/pools";
```
dpwatrous, line 21: Good call - I'm not sure there's any reason to keep AbstractHttpService at this point.

packages/service/src/account/live-account-service.ts
```diff
@@ -0,0 +1,53 @@
...
+        opts?: OperationOptions
+    ): Promise<BatchAccountOutput> {
+        const { subscriptionId, resourceGroupName, batchAccountName } =
+            parseBatchAccountIdInfo(accountResouceId);
+        const cacheManager = getCacheManager();
+        const armBatchClient = createARMBatchClient({
+            baseUrl: getArmUrl(),
+        });
+
+        const cacheKey = `get-${accountResouceId}`;
```
dpwatrous, line 27: How does this cache ever get invalidated? Every Batch account we ever load will be stored in memory until a browser refresh. For accounts this isn't likely to be a huge issue, but it is a memory leak.

packages/react/src/vm-extension/node-vm-ext-list.tsx
wanghoppe: This is only for the FakeAccountService, LiveAccountService will throw 404 error when the account is not found.

packages/react/src/vm-extension/node-vm-ext-list.tsx
wanghoppe: Maybe we can make it a custom hook like `useAccountService` and use `useMemo` inside.

packages/react/src/vm-extension/node-vm-ext-list.tsx
dpwatrous: We should try to be consistent across fake and real implementations here. The `AccountService` interface says it can return `undefined`, so the fact that that's not actually the case for `LiveAccountService` is surprising. I like allowing the `get()` to return `undefined`, but that means we should handle the 404 in `LiveAccountService`.

packages/react/src/vm-extension/node-vm-ext-list.tsx
dpwatrous: Eh, this isn't ever going to be a performance problem - it's just calling a function and reading a property on an object. I'd say let's just use `inject` directly rather than wrapping it. Not all function calls inside a React component need to be hooks.

packages/service/src/account/live-account-service.ts
dpwatrous: We don't necessarily need to add a TTL, and I don't think we need to fully clear the cache either, but leaking every account that was accessed isn't great. Can we do something like an LRU cache with a fixed size for now? That might not scale to all of the ways we might use the cache in the future, but it seems fine for now.

packages/react/src/vm-extension/node-vm-ext-list.tsx
wanghoppe: Sure, I can make `FakeAccountService` to throw if undefined is to return, then we don't need to check the `undefined` everywhere we use `AccountService`

packages/service/src/account/live-account-service.ts
wanghoppe: I just pushed a changes which includes a LRU cache implementation. The LRU cache just record the which key is being used and delete the least recent used cache when the capacity exceeds.

packages/bonito-core/src/cache/cache-manager.ts
```diff
@@ -0,0 +1,98 @@
...
+        const value = await addValue();
+        await this.set(key, value);
+        return value;
+    }
+
+    async get<T>(key: string, defaultValue?: T): Promise<T | undefined> {
+        this.lruCache.recordGet(key);
+        return this.getInternal<T>(key, defaultValue);
+    }
+    abstract getInternal<T>(
```
dpwatrous, line 76: Can these internal methods be marked `protected` so they're only visible to sub-classes?

packages/react/src/vm-extension/node-vm-ext-list.tsx
dpwatrous: Can we reverse that, and return `undefined` when we see a 404? The nice thing about doing that versus throwing an exception is we'll get help from the compiler (and also a null-check is easier than a try/catch). If we really want a version that throws an exception, that's where I'd implement a `getRequired()` method, but I don't think we really need that.

wanghoppe:
Regarding to return undefined when we see a 404. In real case, like in the vm extensions list blade, should we still throw an exception when the account is not found?

PR #2935, opened 2024-09-30
Update ActionForm to support enter key press

skapur12:
This PR is to update the ActionForm so that the user can press the Enter key to submit a form rather than having to click the 'Sign In' button to log in.

This is to fix this CycleCloud bug: https://msazure.visualstudio.com/CycleCloud/_workitems/edit/29267683/

This PR should be merged before this corresponding PR for CycleCloud to make 'Enter' work correctly on the login page:

### DIFF
--- a/packages/bonito-ui/src/components/form/action-form.tsx
+++ b/packages/bonito-ui/src/components/form/action-form.tsx
@@ -105,6 +105,11 @@ export const ActionForm = <V extends FormValues>(
     const [loading, setLoading] = React.useState<boolean>(true);
     const [submitting, setSubmitting] = React.useState<boolean>(false);
 
+    // Store the last typed values to re-populate fields after failure
+    const [lastTypedValues, setLastTypedValues] = React.useState<
+        V | undefined
+    >();
+
     React.useEffect(() => {
         let isMounted = true;
         setLoading(true);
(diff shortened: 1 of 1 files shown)


packages/bonito-ui/src/components/form/action-form.tsx
```diff
@@ -146,6 +151,56 @@ export const ActionForm = <V extends FormValues>(
...
+
+            if (error) {
+                action.logger.error(
+                    "Action execution failed: " +
+                        (error instanceof Error ? error.message : String(error))
+                );
+                if (onExecuteFailed) {
+                    // Restore the form values on failure to avoid clearing, only if lastTypedValues exists
+                    if (lastTypedValues) {
+                        action.form.setValues(lastTypedValues);
```
wanghoppe, line 177: Will the form value be cleared/ changed if the execution failed?

packages/bonito-ui/src/components/form/action-form.tsx
```diff
@@ -146,6 +151,56 @@ export const ActionForm = <V extends FormValues>(
         return <></>;
     }
 
+    // Unified form submission logic
+    const handleSubmit = async () => {
+        setSubmitting(true);
+        try {
+            // Store current form values before execution
+            setLastTypedValues(action.form.values);
```
wanghoppe, line 159: `lastTypedValues` doesn't have to be a react state. Can be achieved by using a local variable
```suggestion
            const lastTypedValues = action.form.values
            ...
            action.form.setValues(lastTypedValues)
```

packages/bonito-ui/src/components/form/action-form.tsx
```diff
@@ -146,6 +151,56 @@ export const ActionForm = <V extends FormValues>(
...
+                onError(e);
+            }
+        } finally {
+            setSubmitting(false);
+        }
+    };
+
+    // Handle the Enter key submission
+    const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+        if (event.key === "Enter" && !submitting) {
```
wanghoppe, line 198: Could consider adding an unit test

PR #2938, opened 2024-10-30
Fixes broken proxy config window

gingi:
Used deprecated Electron `remote` object

### DIFF
--- a/desktop/src/client/proxy/manual-proxy-configuration-window/manual-proxy-configuration-window.html
+++ b/desktop/src/client/proxy/manual-proxy-configuration-window/manual-proxy-configuration-window.html
@@ -5,22 +5,36 @@
     <link href="./manual-proxy-configuration-window.css" rel="stylesheet" type="text/css">
 </head>
 <script>
-    const { ipcRenderer, remote } = require("electron");
-    const initial = remote.getCurrentWindow().currentSettings;
+    const { ipcRenderer } = require("electron");
 
-    document.addEventListener('DOMContentLoaded', function () {
+    if (document.readyState === "loading") {
+        document.addEventListener("DOMContentLoaded", loadValues);
+    } else {
+        loadValues();
+    }
+
+    function loadValues() {
         window.urlEl = document.getElementById("url");
         window.portEl = document.getElementById("port");
         window.usernameEl = document.getElementById("username");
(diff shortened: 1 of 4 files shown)


desktop/src/client/proxy/manual-proxy-configuration-window/manual-proxy-configuration-window.html
```diff
@@ -5,21 +5,24 @@
...
 
-        if (initial) {
-            urlEl.value = initial.url || "";
-            portEl.value = initial.port || "";
-            usernameEl.value = initial.username || "";
-            passwordEl.value = initial.password || "";
-        }
+        configDeferred.then((settings) => {
+            if (settings) {
+                window.urlEl.value = settings.url || "";
```
dpwatrous, line 20: Should we block user interaction or keep the window hidden until this promise resolves? It seems like there could be a race condition here if the window loads, the user starts typing, then the promise resolves and overwrites the values.

desktop/src/client/proxy/manual-proxy-configuration-window/manual-proxy-configuration-window.html
```diff
@@ -5,21 +5,22 @@
...
-    const initial = remote.getCurrentWindow().currentSettings;
+    const { ipcRenderer } = require("electron");
 
-    document.addEventListener('DOMContentLoaded', function () {
-        window.urlEl = document.getElementById("url");
-        window.portEl = document.getElementById("port");
-        window.usernameEl = document.getElementById("username");
-        window.passwordEl = document.getElementById("password");
+    ipcRenderer.invoke("proxy-configuration-requested").then((settings) => {
+        document.addEventListener("DOMContentLoaded", () => {
```
dpwatrous, line 11: We should check `document.readyState` to make sure loading hasn't completed and we've missed the DOMContentLoaded event.

dpwatrous: See: https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete

dpwatrous (changes requested):

dpwatrous (approved):

PR #2941, opened 2024-11-13
Notarization updates

gingi:
* Exits when certificate is invalid
* Uses latest MacOS VM image
* Adds entitlements

### DIFF
--- a/.vsts/darwin/entitlements.plist
+++ b/.vsts/darwin/entitlements.plist
@@ -8,5 +8,11 @@
 	<true/>
 	<key>com.apple.security.cs.disable-library-validation</key>
 	<true/>
+    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
+    <true/>
+    <key>com.apple.security.network.client</key>
+    <true/>
+    <key>com.apple.security.network.server</key>
+    <true/>
 </dict>
 </plist>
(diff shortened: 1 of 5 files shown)


.vsts/darwin/notarize.yml
```diff
@@ -11,7 +11,7 @@ steps:
             "KeyCode" : "CP-401337-Apple",
             "OperationCode" : "MacAppNotarize",
             "Parameters" : {
-              "BundleId": "com.microsoft.azure.BatchExplorer"
+              "BundleId": "com.microsoft.BatchExplorer"
```
dpwatrous, line 14: Is this the correct bundle ID? If so I think we need to update electron-builder.yml as well.

dpwatrous (changes requested):

.vsts/darwin/entitlements.plist
```diff
@@ -8,5 +8,11 @@
 	<true/>
 	<key>com.apple.security.cs.disable-library-validation</key>
 	<true/>
+    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
```
dpwatrous, line 11: Nit: Can we fix the indentation here?

dpwatrous (approved):

PR #2953, opened 2024-12-04
Removes Python from Batch Explorer

gingi:
Removes the Python RPC server, in addition to all Python dependencies. Python SDK code sample templates remain intact.

Other notable changes:

* The Gallery navigation item is removed post deprecation, along with Gallery-related settings
* Folder uploading is fixed to honor local directory structure
* Fixes UI error alerts, which were previously empty because the error response was parsed incorrectly
    Batch_Explorer


gingi:
Fixes #2954, #2955, #2956, #2957

desktop/src/@batch-flask/core/server-error/server-error.ts
```diff
@@ -124,7 +127,13 @@ export class ServerError {
     }
 
     public static fromARM(response: HttpErrorResponse): ServerError {
-        const error = response.error || {};
+        // KLUDGE: Angular's request nests the error inside an error object
+        let error;
+        if (response.error?.error) {
+            error = response.error.error;
+        } else {
+            error = response.error;
```
dpwatrous, line 135: Should this be `response.error || {}` to preserve the previous behavior? Also `response.error.error || {}` above. If `error` ends up as null or undefined, line 139 below will error.

desktop/src/app/components/data/shared/file-group-sas/file-group-sas.component.ts
```diff
@@ -80,9 +80,6 @@ export class FileGroupSasComponent implements ControlValueAccessor, OnChanges, O
             this.fileGroupsData.params = {
                 storageAccountId,
             };
-            this.fileGroupsData.setOptions({
-                prefix: Constants.ncjFileGroupPrefix,
```
dpwatrous, line 84: Should this be legacyFileGroupPrefix?

dpwatrous (approved):

PR #2973, opened 2025-02-14
Fixes proxy issue with browser auth

gingi:
See [this `msal-node` issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6527).

### DIFF
--- a/desktop/src/client/core/aad/proxy-network-client.ts
+++ b/desktop/src/client/core/aad/proxy-network-client.ts
@@ -1,0 +1,55 @@
+import type { INetworkModule, NetworkRequestOptions, NetworkResponse } from "@azure/msal-node";
+import * as HttpsProxyAgent from "https-proxy-agent";
+import fetch from "node-fetch";
+
+/**
+ * Placeholder for msal-node's network module which uses node-fetch to support
+ * HTTP proxy configurations with authorization
+ *
+ * @see https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6527#issuecomment-2077953882
+ */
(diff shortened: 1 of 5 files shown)


desktop/src/client/core/aad/proxy-network-client.ts
```diff
@@ -0,0 +1,55 @@
...
+            });
+
+            return {
+                headers: headersObj,
+                body: data,
+                status: response.status,
+            };
+        } catch (err) {
+            console.error("CustomRequest", err);
+            throw new Error("Custom request error");
```
dpwatrous, line 52: Can we just re-throw the original exception?

desktop/src/client/core/aad/proxy-network-client.ts
```diff
@@ -0,0 +1,55 @@
...
+    sendGetRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<NetworkResponse<T>> {
+        return this.sendRequestAsync(url, "GET", options);
+    }
+    sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<NetworkResponse<T>> {
+        return this.sendRequestAsync(url, "POST", options);
+    }
+
+    private async sendRequestAsync<T>(
+        url: string,
+        method: "GET" | "POST",
```
dpwatrous, line 26: What about other HTTP verbs? (PUT/PATCH/DELETE/etc)

gingi: We don't need to handle other verbs because the Entra handshake only involves `GET` and `POST`. This network client is only being used for authentication.

